PR #3, opened 2017-10-16
aligning dependencies with jboss-parent and minor simplifications

shawkins:


komodo-core-tests/pom.xml
```diff
@@ -58,7 +58,7 @@
 
 		<dependency>
 			<groupId>org.mockito</groupId>
-			<artifactId>mockito-core</artifactId>
+			<artifactId>mockito-all</artifactId>
```
phantomjinx: In the commit message, can you specify the reason you need all of mockito rather than just the core. Running most of the teiid unit tests did not seem to require anything more than mockito-core. Either way, if you need all of it, no problem - just specify reason in commit.

komodo-modeshape/pom.xml
```diff
@@ -23,5 +23,13 @@
...
 		</plugins>
 	</build>
+	
+	<dependencies>
+		<dependency>
+			<groupId>log4j</groupId>
+			<artifactId>log4j</artifactId>
+			<scope>compile</scope>
+		</dependency>
+	</dependencies>
```
phantomjinx: Can you provide a reason as to why this was required, please?

komodo-parent/pom.xml
```diff
@@ -1,11 +1,11 @@
...
-		<groupId>org.jboss.integration-platform</groupId>
-		<artifactId>jboss-integration-platform-parent</artifactId>
-		<version>6.0.1.Final</version>
-		<relativePath></relativePath>
-	</parent>
+      <groupId>org.jboss</groupId>
+      <artifactId>jboss-parent</artifactId>
+      <version>19</version>
+      <relativePath></relativePath>
+    </parent>
```
phantomjinx: I realise teiid was under jboss-parent but komodo was asked to be put under integration-platform-parent by QE. Is there a reason this needs to be under one or the other?

komodo-parent/pom.xml
```diff
@@ -183,6 +175,14 @@
...
 				<scope>import</scope>
 			</dependency>
+			<!--IP-Platform BOM -->
+			<dependency>
+				<groupId>org.jboss.integration-platform</groupId>
+				<artifactId>jboss-integration-platform-bom</artifactId>
+				<version>${version.ip-bom}</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
```
phantomjinx: Why has this moved in the pom?

komodo-parent/pom.xml
```diff
@@ -536,49 +536,12 @@
...
-				<version>${version.org.slf4j}</version>
-			</dependency>
-
-			<!--Inherited from BOM, but changes the default scope to "test" -->
-			<dependency>
-				<groupId>log4j</groupId>
-				<artifactId>log4j</artifactId>
-				<version>${version.log4j}</version>
-			</dependency>
-
```
phantomjinx: I assume there is a reason these have been removed. Can you specify please?

komodo-spi/src/main/java/org/komodo/spi/utils/KeyInValueHashMap.java
```diff
@@ -21,22 +21,18 @@
...
- * own keys. More memory efficient than storing separate references
- * to the keys and values.
+ * own keys.
  *
  * @param <K> key
  * @param <V> value
  */
 @SuppressWarnings( "unchecked" )
-public class KeyInValueHashMap<K, V> extends AbstractMap<K, V> {
+public class KeyInValueHashMap<K, V> extends LinkedHashMap<K, V> {
```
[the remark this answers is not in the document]
shawkins: > This implementation was meant to improve memory footprint.

Which it did not.  A HashSet simply wraps a HashMap, then the usage of a EntryWrapper adds another Object/reference into the picture.  If you profile the retained heap size of this vs. the old implementation you will see that the old was larger.

> By making it a sorted map, it raises the memory profile for an implementation that only occasionally requires the keys to be sorted.

Linked does not imply sorted.  Linked only maintains the insertion order, not the natural ordering.  You can certainly change that to just a HashMap, as LinkedHashMap does add a minor amount of overhead to each entry (2 references).  However what I've seen a number of times on the Teiid side are unit tests or other places in the code that inadvertently rely on the ordering of the hashing, only to have that change in the next java version.

komodo-parent/pom.xml
shawkins: I would assume that you do not want to generally make log4j a compile dependency correct?

komodo-modeshape/pom.xml
shawkins: This is paired with the other log4j change.  Since it is no longer a general compile time dependency, it needs to be added specifically for modeshape.

komodo-core-tests/pom.xml
shawkins: I'm not sure the commit message is the right place to cover every change in the commit set.  This is to align with jboss-parent, which uses all rather than core.

komodo-parent/pom.xml
shawkins: To align with later Teiid/WildFly community.  With the assumption that we are moving towards swarm.  Also the build was failing with an httpclient related error in the s3 resource-adapter.  This appeared to be the right direction to resolve that.  Another approach would be to look at using a 7.0.x version of the integration-platform-parent, but since that seems to diverge from WildFly this seemed better.

komodo-parent/pom.xml
shawkins: To resolve the httpclient build issue.

komodo-core-tests/pom.xml
phantomjinx: The commit message is absolutely the place to record such changes. Its the only log that remains with the source code. It should never be necessary to refer to an alternative source to understand decisions for a specific code change. Certainly, common coding standards should require that some description be added to a commit message.

komodo-parent/pom.xml
phantomjinx: Cool. This kind of design decision needs to be logged in the commit message, please.

komodo-parent/pom.xml
phantomjinx: Understand. Again, a quick line in the commit message, please?

komodo-spi/src/main/java/org/komodo/spi/utils/KeyInValueHashMap.java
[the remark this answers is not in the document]
phantomjinx: You seem to have confused my comments. I am not suggesting that the 'new' version is larger/smaller than the 'old' version. The memory footprint comment refers to specifically using a LinkedHashMap which, as you say, includes additional overhead, compared to an AbstractMap.

'sorted' was obviously the incorrect word.

Again, if you have encountered instances where the solution is better, please say so in the commit message. Such comments preclude questioning and misunderstanding.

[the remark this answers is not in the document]
shawkins: Yes, the confusion stemmed from the usage of sorted.

> Such comments preclude questioning and misunderstanding.

I'm honestly not trying to shirk comments.  Since the web.xml and this appeared to be minor changes upon initial review I simply included in this pull request - I will separate such commits for clarity.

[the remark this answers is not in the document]
phantomjinx: Thanks very much. It is much appreciated.

PR #16, opened 2017-12-15
TEIIDTOOLS-283: 

rareddy:
1) Upgraded Teiid version to 10.0.0.Final
2) Also upgraded the WF-Swarm to reflect the version of the Teiid.
3) Changed the maven group ids to reflect the new version
4) Added a dependency on https://github.com/teiid/kubernetes-service-catalog-client project, which is library required to interact with service catalog (this needs to put in maven central too)
5) Added code to create a docker image based on F8 docker plug-in. This is helpful in releasing a final image, which is tested with the template image at https://github.com/rareddy/osb-data-services This project will be merged with @shawkins work on community-based images. These would be community deliverables.
6) The above template uses oauth-proxy to let user login into OpenShift, which is a http proxy
7) Based on headers from the oauth-proxy in the Komodo a AuthProxy has been added to inject the access-token an user name into a known context that is accessible to rest of the modules. At a later point, we will consider building a full security context based on these.
8) Also added a PG image to the template, which can be used as storage for modeshape.
With all above in place, if there is a Posgresql database that has been provisioned to the project, a call to Komodo's rest api /v1/metadata/connections will contact the service broker and get list of services, and then, if required it will bind to the service (if not already bound), and will gather the properties required for to make connection from the OpenShift Secrets and using the Teiid's Admin API will create a connection, and then return the resulting connections available in the Komodo-Teiid engine.

### DIFF
--- a/komodo-core/src/main/resources/org/komodo/core/repository/local-repository-config.json
+++ b/komodo-core/src/main/resources/org/komodo/core/repository/local-repository-config.json
@@ -20,15 +20,15 @@
             "type" : "db",
             "connectionUrl": "${komodo.connectionUrl}",
             "driver" : "${komodo.connectionDriver}",
-            "username" : "komodo",
-            "password" : "komodo"
+            "username" : "${komodo.user}",
+            "password" : "${komodo.password}"
         },
         "binaryStorage" : {
           "type"  : "database",
           "driverClass" : "${komodo.connectionDriver}",
           "url" : "${komodo.binaryStoreUrl}",
-          "username" : "komodo",
-          "password" : "komodo"
+          "username" : "${komodo.user}",
+          "password" : "${komodo.password}"
         },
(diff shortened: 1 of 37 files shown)


rareddy:
@phantomjinx Ok to review. Make any changes necessary and pull in. 
1) I do not think PersistenceType is no longer needed with this. You can remove or refactor to use.
2) Feel free to change Rest API and its methods, it mostly a cut n paste I did. No repository stuff there, if needs to be add in. I have no idea about that side so I left it for you amend as you see fit with rest of it.
3) All unit tests including arquillian tests were passing, I did not write for new ones because I need some kind of mock for service catalog stuff to do that.
4) most of the classes are boiler plate code for adding new api calls.

phantomjinx:
Merged with additional changes to appear in further PRs

PR #25, opened 2018-02-01
Dataservice multi-view support

mdrillin:
TEIIDTOOLS-323
(This PR has a companion beetle-studio PR)
- modifies DataserviceService rest method to accept multiple tables, which are then used to generate the service views
- modifies Dataservice getServiceView(s) to return array of view names rather than single view
- fixed associated tests


phantomjinx:
Looks good

PR #28, opened 2018-02-21
TTOOLS-337 Support for service catalog sources

mdrillin:
Note: this PR has a corresponding beetle-studio PR
- Added methods to KomodoConnectionService to create/update connections using an openshift serviceCatalogDataSource.  The methods ensure that the provided serviceCatalogDataSource is 'bound' - then the wildfly source is used to create/update a corresponding source in the komodo repo.
- Other attribute classes, etc in support of the new methods
- Added ServiceCatalogDataSourceSerializer and registered with KomodoJsonMarshaller - to resolve issue with associated responses.
- Couple fixes in TeiidSwarmMetadataInstance for incorrect property names


rareddy:
@mdrillin was there a reason why you changed the MySQL connection property names? I believe I got those names from the MySQL template on OpenShift. Repository based connection creation/saving I am much less familiar with how that works. 

However, I thought you already have the connection famework from before that can be used for saving service catalog based connections, as there is no difference in service catalog based connections vs ones created by collecting the user/pass/url once the service has been provisioned. That way I think that what you have here verbose or duplicate, instead the previous connection API should have been modified to cater to service catalog based connections.

mdrillin:
I was testing with a MySQL source and changed the property names because the original names for MySQL did not work.  Yes the added methods could have replaced the originals, assuming the original approach is no longer needed.

server/komodo-rest/src/main/java/org/komodo/rest/TeiidSwarmMetadataInstance.java
```diff
@@ -78,6 +78,8 @@ public TeiidSwarmMetadataInstance(TeiidConnectionProvider connectionProvider) {
 							scd.setType(type.name());
 							ServiceBinding binding = getServiceBinding(svc);
 							if (binding == null) {
+								scd.setBound(false);
+							} else {
 								scd.setBound(true);
```
phantomjinx: Does this not change this logic to the opposite. Is that expected?

server/komodo-rest/src/main/java/org/komodo/rest/TeiidSwarmMetadataInstance.java
```diff
@@ -260,9 +262,9 @@ private void createDataSource(String name, SourceType type, Map<String, String>
 			 */
 			Properties props = new Properties();
 			props.setProperty("connection-url", "jdbc:mysql://"+map.get("DATABASE_SERVICE_NAME")+":3306/" 
-					+ map.get("database-name"));
-			props.setProperty("user-name", map.get("database-user"));
-			props.setProperty("password", map.get("database-password"));
+					+ map.get("database_name"));
+			props.setProperty("user-name", map.get("username"));
+			props.setProperty("password", map.get("password"));
```
phantomjinx: Are these property names not guaranteed to exist then, ie. openshift service dependent?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
```diff
@@ -377,6 +387,129 @@ public Response createConnection( final @Context HttpHeaders headers,
...
+            Collection<ServiceCatalogDataSource> dataSources = getMetadataInstance().getServiceCatalogSources();
+			for(ServiceCatalogDataSource ds: dataSources) {
+				if(ds.getName().equals(rcAttr.getServiceCatalogSource())) {
+					serviceCatalogSource = ds;
+					break;
+				}
+			}
+			// If catalogSource is not found, exit with error
+			if(serviceCatalogSource == null) {
+                return createErrorResponseWithForbidden(mediaTypes, RelationalMessages.Error.CONNECTION_SERVICE_CATALOG_SOURCE_DNE_ERROR);
```
phantomjinx: Really really minor point -> can this method be space/tab formatted?

server/komodo-rest/src/main/java/org/komodo/rest/TeiidSwarmMetadataInstance.java
rareddy: yes, that was bug. I also have in my changes.

rareddy:
I suspect we may be running into multiple different templates perhaps, I should think more about having multiple properties names, for now, it is OK with change.

As per needed altering the original code, I am suggesting to modify such that it handles both, not have two seperate APIs. From Steve's comments about connection wizrds we may need to have both at a later stage.

server/komodo-rest/src/main/java/org/komodo/rest/TeiidSwarmMetadataInstance.java
rareddy: I would say as soon as we integrated with Service Catalog we were OpenShift dependent, but IMO there is nothing wrong with that. What we are seeing is an effect of multiple different templates for MySQL. We need to think little more about how to handle it.

I would even go further and say that we should tailor more knowing the platform and remove any APIs/code that is no longer needed. I am not saying there is, but it was there..

phantomjinx: @rareddy you misunderstand "openshift service dependent". I mean in the context of different implementations of mysql instances or postgresql instances in Openshift not independent of Openshift. Your comment goes on to illustrate my question in more detail.

Certainly need to think a little more as there is no guarantee that any one template is going to conform to a particular set of property names. For all we know "user" could be "username" or "uname" or "uow" or even "namespace" or "blah" (probably not the latter). That's going to require some serious handling!

server/komodo-rest/src/main/java/org/komodo/rest/TeiidSwarmMetadataInstance.java
mdrillin: The original logic did not look correct.  Seems that if the binding is not found, then 'bound' should be false?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
mdrillin: yes

mdrillin:
What are the 'must fix' to get this merged?
- I'm not in agreement with modifying the rest call at this point with "handling both" way of creating connections.  I don't have a clear understanding of what the "other" way is, so it does not make sense to me to add at this point.
- I am unclear based on the previous comments what needs modification for merge acceptance.

phantomjinx:
@mdrillin 
From my perspective, I suggest modifying TeiidSwarmMetadataInstance@L265 to support both the original user/password properties and your new versions. This would at least allow support for 2 different database-esque data source templates. Once that's done, I'm good with merging.

rareddy:
There is no must fix IMO. Just verbose APIs. The suggestion was to combine the current code with the previous connection API. Once we pass this, there would be no incentive to fix this ever. Just having two different APIs doing same things seems confusing IMO that's all.

mdrillin:
@phantomjinx I will assign a JIRA to you for the TeiidSwarmMetadataInstance work as I'm not sure what you mean

@rareddy I would prefer to remove the old methods rather than try to merge differing functionality

mdrillin:
Made the requested changes to the api

PR #29, opened 2018-02-27
TEIIDTOOLS-317 Add Travis CI

asmigala:


rareddy:
can you update the pull with JIRA #? Also docker.tag could be just $DOCKER_USER_NAME ?

asmigala:
Updated the commit message with the correct JIRA number.

Regarding the docker tag, I'm not sure what exactly you mean. Should I refer directly to $DOCKER_USERNAME in the pom? That does not sound right to me, as it requires that the user exports such variable to be able to build the docker image locally.

rareddy:
I believe you can define default user as Teiid and ability to pass the system property.

asmigala:
I added a separate property for docker username as well as for the resulting image name used in the tag. They default to "teiid/komodo" and can be overriden by passing the property to the maven build.

Hope that is satisfactory.

rareddy:
That looks good, but I am confused how you are setting your docker name. Are you pushing to the "teiid" docker hub or that what you need to the credentials for or you want to setup another dockerhub account? I thought you were going for another one. If not i can provide the credentials for teiid docker hub.

asmigala:
In my initial testing, I set the docker username to "asmigala", so that it would get deployed into my docker hub account. For the PR, I removed that part, because the intent is that the images will be built from this repo (teiid/teiid-komodo) and pushed into the "official" community docker hub account, i.e. teiid/komodo.

rareddy:
Thanks @asmigala

PR #33, opened 2018-03-14
TEIIDTOOLS-357

mdrillin:
Changes to the rest interfaces to support management of connection VDBs
- Add KomodoConnectionService method which deploys a VDB for a given connection.
- Adds KomodoConnectionService method which returns VDBStatus for all workspace Connection objects.
- Adds KomodoDataserviceService method which returns VDBStatus for all workspace Dataservice objects.
- modified RestMetadataVdbStatusVdb and serializer to support instance when no VDB exists for the workspace object; in that case returns empty status.
- remove unused method in KomodoDataserviceService and other minor removals.

### DIFF
--- a/server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
+++ b/server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
@@ -21,6 +21,7 @@
  */
 package org.komodo.rest.service;
 
+import static org.komodo.rest.relational.RelationalMessages.Error.CONNECTION_SERVICE_DEPLOY_CONNECTION_VDB_ERROR;
 import static org.komodo.rest.relational.RelationalMessages.Error.CONNECTION_SERVICE_NAME_EXISTS;
 import static org.komodo.rest.relational.RelationalMessages.Error.CONNECTION_SERVICE_NAME_VALIDATION_ERROR;
 import static org.komodo.rest.relational.RelationalMessages.Error.VDB_DATA_SOURCE_NAME_EXISTS;
@@ -47,9 +48,13 @@ import javax.ws.rs.core.UriInfo;
 
 import org.komodo.core.KEngine;
 import org.komodo.core.repository.ObjectImpl;
+import org.komodo.relational.DeployStatus;
 import org.komodo.relational.connection.Connection;
+import org.komodo.relational.model.Model;
+import org.komodo.relational.vdb.ModelSource;
 import org.komodo.relational.vdb.Vdb;
 import org.komodo.relational.workspace.WorkspaceManager;
+import org.komodo.rest.KRestEntity;
 import org.komodo.rest.KomodoRestException;
 import org.komodo.rest.KomodoRestV1Application.V1Constants;
 import org.komodo.rest.KomodoService;
@@ -59,6 +64,8 @@ import org.komodo.rest.relational.connection.RestConnection;
 import org.komodo.rest.relational.json.KomodoJsonMarshaller;
 import org.komodo.rest.relational.request.KomodoConnectionAttributes;
 import org.komodo.rest.relational.response.KomodoStatusObject;
+import org.komodo.rest.relational.response.RestObjectVdbStatus;
+import org.komodo.rest.relational.response.metadata.RestMetadataVdbStatusVdb;
 import org.komodo.servicecatalog.TeiidOpenShiftClient;
 import org.komodo.spi.KException;
 import org.komodo.spi.constants.StringConstants;
@@ -70,6 +77,7 @@ import org.komodo.spi.repository.Repository.UnitOfWork;
 import org.komodo.spi.repository.Repository.UnitOfWork.State;
 import org.komodo.spi.runtime.ServiceCatalogDataSource;
 import org.komodo.spi.runtime.TeiidDataSource;
+import org.komodo.spi.runtime.TeiidVdb;
 import org.komodo.utils.StringUtils;
 
 import io.swagger.annotations.Api;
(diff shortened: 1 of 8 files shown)


rareddy:
@mdrillin Why is this any different from any VDB deployment that REST API may already have?

mdrillin:
It doesnt change any existing VDB deployment.  In the case of connection VDB deployment it encapsulates a lot of the work that the client was required to do previously.

rareddy:
@mdrillin OK, I would have preferred to change the current endpoint with additional info to denote the "connection" based VDB in the parameters and handled that way. More endpoints more code, more to manage.

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
```diff
@@ -835,6 +849,204 @@ public Response validateConnectionName( final @Context HttpHeaders headers,
...
+     */
+    @POST
+    @Path( V1Constants.CONNECTION_DEPLOY_VDB_SEGMENT + StringConstants.FORWARD_SLASH + V1Constants.CONNECTION_PLACEHOLDER )
+    @Produces( { MediaType.APPLICATION_JSON } )
+    @ApiOperation( value = "Deploy a VDB for the workspace connection" )
+    @ApiResponses(value = {
+            @ApiResponse(code = 406, message = "Only JSON is returned by this operation"),
+            @ApiResponse(code = 403, message = "An error has occurred.")
+        })
+    public Response deployVdbForConnection( final @Context HttpHeaders headers,
```
phantomjinx: I can't help feeling that we are mixing and confusing our resources. Up until now:
* KomodoVdbService has dealt with managing vdbs in the repository;
* KomodoMetadataService has dealt with the teiid server and metadata deployment;
* KomodoConnectionService has dealt with manager connections in the repository.

This seems to now be merging all three. I can see the value of this but I would propose separating this into a different service on its own, maybe KomodoCompositeService or KomodoManagementService?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
```diff
@@ -835,6 +849,204 @@ public Response validateConnectionName( final @Context HttpHeaders headers,
...
+            final Vdb existingVdb = findVdb( uow, vdbName );
+
+            if ( existingVdb != null ) {
+                mgr.delete(uow, existingVdb);
+            }
+            
+            // Create new VDB
+            String vdbPath = repoPath + "/" + vdbName;
+            final Vdb vdb = getWorkspaceManager(uow).createVdb( uow, null, vdbName, vdbPath );
+            vdb.setDescription(uow, "Vdb for connection "+connectionName);
```
phantomjinx: This code should be shared with KomodoVdbService so that both are doing the same job. Otherwise, such duplication will produce greater effort to maintain.

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
```diff
@@ -835,6 +849,204 @@ public Response validateConnectionName( final @Context HttpHeaders headers,
...
+            // Add model source to the model
+            ModelSource modelSource = model.addSource(uow, connectionName);
+            modelSource.setJndiName(uow, jndiName);
+            modelSource.setTranslatorName(uow, driverName);
+            
+            // Deploy the VDB
+            DeployStatus deployStatus = vdb.deploy(uow);
+            
+            // Await the deployment to end
+            Thread.sleep(DEPLOYMENT_WAIT_TIME);
```
phantomjinx: This code should be shared with KomodoMetadataService rather than duplicated for the same reason.

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
```diff
@@ -835,6 +849,204 @@ public Response validateConnectionName( final @Context HttpHeaders headers,
...
+    @Path(V1Constants.VDB_STATUS_SEGMENT)
+    @Produces( MediaType.APPLICATION_JSON )
+    @ApiOperation(value = "Display the VDBStatus of all connections in the workspace",
+                  response = RestObjectVdbStatus.class)
+    @ApiResponses(value = {
+        @ApiResponse(code = 403, message = "An error has occurred.")
+    })
+    public Response getVdbStatuses(final @Context HttpHeaders headers,
+                                   final @Context UriInfo uriInfo) throws KomodoRestException {
+
```
phantomjinx: This should be moved to KomodoMetadataService and made generic for any named artifact, ie. connection or dataservice.

This ends up being really confusing for the API. We would have:
* a vdb resource that provides GET information of vdbs
* a connection resource that provides GET information on connections but also now provides GET information on connection vdbs too

Do we really want to expose the vehicle that a connection needs to be deployed in, ie. a vdb, as it is a rather confusing implementation-only concept?

This maybe should be getDeployedConnectionStatuses (although something more pithy and clear would be better).

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoDataserviceService.java
```diff
@@ -371,6 +377,78 @@ public Response getDataservice( final @Context HttpHeaders headers,
...
+    @GET
+    @Path(V1Constants.VDB_STATUS_SEGMENT)
+    @Produces( MediaType.APPLICATION_JSON )
+    @ApiOperation(value = "Display the VDBStatus of all dataservices in the workspace",
+                  response = RestObjectVdbStatus.class)
+    @ApiResponses(value = {
+        @ApiResponse(code = 403, message = "An error has occurred.")
+    })
+    public Response getVdbStatuses(final @Context HttpHeaders headers,
+                                   final @Context UriInfo uriInfo) throws KomodoRestException {
```
phantomjinx: see above.

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoConnectionService.java
rareddy: That tells me that there is existing functionality in REST API, but it is not modular enough to re-use in certain ways. I suggest we look as to how to re-use or combine multiple calls to achieve same.

mdrillin:
Some very good points made here about the overall REST interface design.  I will close this PR and the corresponding UI PR, then open a new JIRA to review the overall REST interface design.  We need an interface that is generic and modular, yet allows clients to perform meaningful tasks without chaining together many calls.

PR #37, opened 2018-03-26
Changes made to better enable UI support for publishing of services

phantomjinx:
Please see commit messages and hopefully they should provide the necessary explanations for changes made.

This should be merged alongside [beetle-studio-PR-113](https://github.com/teiid/beetle-studio/pull/113)


server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
```diff
@@ -3576,17 +3581,97 @@ public Response publishVirtualization(final @Context HttpHeaders headers, final
...
-            Repository repo = this.kengine.getDefaultRepository();
-            uow = createTransaction(principal, "publish", true); //$NON-NLS-1$
+            uow = createTransaction(principal, "publish-init", true); //$NON-NLS-1$
             Vdb vdb = findVdb(uow, attributes.getName());
+            if (vdb == null) {
+                //
+                // We don't have a vdb so maybe we have a dataservice instead
+                // Find the dataservice's vdb to publish.
+                //
+                Dataservice dataservice = findDataservice(uow, attributes.getName());
```
rareddy: What is a DataService in this context of the Komodo application? Are there different ways of VDB availability?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
```diff
@@ -3576,17 +3581,97 @@ public Response publishVirtualization(final @Context HttpHeaders headers, final
...
+            final AuthToken token = getAuthenticationToken();
+
+            //
+            // This publishing could take 30+ seconds to return hence the need to
+            // check it then return before the http request actually times out.
+            //
+            // Waits up to 10 seconds to see if the publishing returns (and probably
+            // has failed!) then returns an indication that things are underway.
+            //
+            ExecutorService threadService = Executors.newFixedThreadPool(1);
```
rareddy: Why would publishing timeout in 30 secs? I did not see this behavior when I was using swagger API. Timeout IMO is typically set by the calling client on their expectations. The publishing had sync part and async part, you are adding another layer on top of it which I do think is correct.

If for some reason you need to return in 10secs, we could have modified the publishing API to be totally async with one more state, maybe like "submitted" and encapsulate this behavior to return the call immediately (not wait for pom generation, initial build)

server/komodo-rest/src/main/java/org/komodo/rest/relational/response/virtualization/RestVirtualizationStatus.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.Date;
+
+import org.komodo.rest.AbstractKEntity;
+import org.komodo.servicecatalog.BuildStatus;
+import org.komodo.spi.KException;
+import org.komodo.spi.repository.KomodoObject;
+import org.komodo.spi.repository.Repository.UnitOfWork;
+import org.komodo.utils.ArgCheck;
+
+public class RestVirtualizationStatus extends AbstractKEntity {
```
rareddy: AbstractKEntity? I thought you said if we are not storing in Komodo repo we do not need it?

server/komodo-rest/src/main/java/org/komodo/rest/relational/json/VirtualizationStatusSerializer.java
```diff
@@ -0,0 +1,116 @@
...
+import org.komodo.rest.Messages;
+import org.komodo.rest.relational.response.virtualization.RestVirtualizationStatus;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
+/**
+ * A GSON serializer/deserializer for {@status RestTeiidVdbStatusVdb}s.
+ */
+public final class VirtualizationStatusSerializer extends TypeAdapter< RestVirtualizationStatus > {
```
rareddy: Same here as KEntity, if not stored in the repository there is no requirement for custom serializer?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
phantomjinx: Beetle studio, like DSB, interacts with teiid-komodo at the 'dataservice' layer (you may recall I started a discussion on IRC a week or so ago concering the need for dataservices). Therefore, the vdb created by beetle-studio/teiid-komodo when the user has completed the GUI wizard is in fact wrapped in a dataservice. Now, it is possible to extract the 'service-vdb' (the main vdb of the dataservice since the latter can contain others for import), however, making beetle-studio do that work is unnecessary given that the dataservice API remains.

At some point, this can be removed along with the entire dataservice framework if that is felt to be the way forward but for now that is out of scope for this sprint. Therefore, allowing this method to take the name of a vdb or a dataservice covers all options and reduces the amount of business logic necessary in beetle-studio.

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
phantomjinx: 1. Publishing does not timeout in 30 seconds, the http request does. Please see:
* https://seancoates.com/blogs/api-gateway-timeout-workaround/
* https://stackoverflow.com/questions/13582409/http-client-timeout-and-server-timeout

2. As noted in the commit messages, the publishing operation up to adding-to-queue (when the new thread kicks in) takes 28-45 seconds on my computer. It is possible that your computer has been running on the low end of this range so have avoided it.

3. This timeout IS set by the calling client, ie. the browser and 30 seconds is plenty long enough to keep a TCP connection open.

4. The threading is necessary as the sync part is taking too long and returning a 504 Gateway Timeout error when it shoud return a 200 response.

5. Whether this is another layer is irrelevant. It is necessary for any operation that on the system will take a long time. For example, if exporting or connection creation in the future take longer than 30 seconds then they too will require a threading modification.

6. The operation did NOT need to return in 10 seconds and, as stated in the commit messages, provides a means of ensuring that the if the publishing returns prematurely due to an error, that error can be immediately returned via the direct response rather than awaiting the polling. The 10 seconds can be reduced but upon testing seemed a useful middle-of-the-road value.

7. Your suggestion of further modification of the publishing API to be totally async is a good one. However, within the scope of this set of commits, I felt it was one step too far. Certainly, a follow-on to introduce additional states would be beneficial, especially as I have noted the following (but not actually witnessed it):
 * User initiates publish
 * Publish request goes to komodo and kicks of the 'publishing build' thread
 * The request waits 10 seconds for any error then returns a response of publishing initiated
 * The UI returns a successful message to say publishing under way and the icon starts to spin
 * Meanwhile, the building of the content is slow and takes 45 seconds rather than ~32 and at that point
 * The UI polls komodo and notes there is nothing in the queue so changes the publishing icon back to "Not Published" (not spinning)
 * The UI polls again, finds something on the queue and changes the icon back to "publishing" (spinning)

To reiterate I have not seen this in testing as my build takes about ~32 seconds which means its possible that the polling never kicks in before the publishing service joins the queue. Therefore, putting everything on a permanent queue would be better to avoid this potential.

server/komodo-rest/src/main/java/org/komodo/rest/relational/response/virtualization/RestVirtualizationStatus.java
phantomjinx: Nothing is being stored in the Komodo repo. Please see the hierarchy of those classes that inherit from KRestEntity.

Objects that do get stored in the repository tend to inherit from RestBasicEntity since this class included methods such as setId(), setDataPath() etc.

## Reason for Change
* getVirtualizations() returns a RestVirtualizationState (renamed from RestBasicState)
* With no serializer the standard json for the KRestEntity is returned in an http response, ie.
```json
{
  tuples: {
    buildName: 'blah'
    vdbName: 'blah'
    ....
  }
}
```

* In beetle-studio ts-script allows a very quick mechanism for turning json structures into javascript objects ('classes'). As such, a virtualization model has been added for tracking the getVirtualization() responses.
* With the extra level of tuples in the json response, it becomes harder to change the response into a virtualization model. Can be done but it is far easier with a serializer on the komodo side.
* The serializer removes the extra level and returns the json as:
```json
[
  buildName: 'blah'
  vdbName: 'blah'
  ....
],
[
....
]
```

* The response becomes more useful and easy to convert with javascript.

## Conclusion
1. A serializer was required due to a custom conversion;
2. The RestVirtualizationStatus is NOT being stored in the repository;

server/komodo-rest/src/main/java/org/komodo/rest/relational/json/VirtualizationStatusSerializer.java
phantomjinx: See above.

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
```diff
@@ -3579,13 +3579,28 @@ public Response publishVirtualization(final @Context HttpHeaders headers, final
             Repository repo = this.kengine.getDefaultRepository();
             uow = createTransaction(principal, "publish", true); //$NON-NLS-1$
             Vdb vdb = findVdb(uow, attributes.getName());
+            if (vdb == null) {
+                //
+                // We don't have a vdb so maybe we have a dataservice instead
+                // Find the dataservice's vdb to publish.
+                //
+                Dataservice dataservice = findDataservice(uow, attributes.getName());
```
mdrillin: In the case of user having a same name vdb and dataservice in their repo, maybe the user intended to publish the dataservice.  There is no way to do that with this method, since it looks for a vdb with that name first.  Maybe also need to provide the type in the request to alleviate that?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
rareddy: Unfortunately, I was not part of the IRC discussion. I see your point. I strongly encourage the removal of Data Service to bring clarity to API

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
rareddy: Can you create a JIRA for this for full async mode? I will see if I can fix that. What is your polling frequency?

server/komodo-rest/src/main/java/org/komodo/rest/relational/response/virtualization/RestVirtualizationStatus.java
rareddy: I wish your default "properties" value encoding to JSON converted to an array instead of JSON object to avoid the custom serializers. I thought I added the values to properties, not tuples, and that could be the distinction between uses of when to use which one in the AbstractEntity classes.

server/komodo-rest/src/main/java/org/komodo/rest/relational/json/VirtualizationStatusSerializer.java
rareddy: got it, thanks confused over KRestEnity over KAbstractEntiry. Maybe a better name is KRespositoryEntity and ReponseEnitity, WDYT?

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
rareddy: There should be no DataService in Beetle IMO, we need weed it out as that is what we did in DSB and I do not think that will fit here. It causes more confusion than anything else.

mdrillin: The reason we created dataservice object initially was to encapsulate everything needed to make the service portable to another server.  That was well before our openshift service strategy.  So yeah, we can probably move away from using it at this time.  This will require some redesign and is a topic for future sprint(s).

rareddy: @mdrillin Add a JIRA, then we can identify first what we can put on the chopping block

mdrillin: TEIIDTOOLS-399

server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
phantomjinx: see [TEIIDTOOLS-400](https://issues.jboss.org/browse/TEIIDTOOLS-400).

Polling occurs every 5 seconds.

server/komodo-rest/src/main/java/org/komodo/rest/relational/json/VirtualizationStatusSerializer.java
phantomjinx: Certainly can consider this.

phantomjinx:
Think all questions answered and follow-on issues logged so will merge this. If anything further is required then can rework in new PRs.

PR #39, opened 2018-04-04
TEIIDTOOLS-371 Rest services for connection schema management

mdrillin:
- Reworks the get connection services to allow for retrieval of the schema status along with a connection.  The services now return a 'ConnectionSummary'.  With no params specified, the summary will just include the Connection.  The user can also provide 'include-connection' and 'include-schema-status' parameters to get just the connections, just the connection schema status, or both the connection and the schema status.
- Adds a post service to initiate refresh of the connection schema.  The deploys/re-deploys the connection vdb.
This PR has a companion PR in beetle-studio.  The should be merged together.


server/komodo-rest/src/main/java/org/komodo/rest/relational/response/RestConnectionSummary.java
```diff
@@ -0,0 +1,96 @@
...
+import java.net.URI;
+
+import org.komodo.rest.RestBasicEntity;
+import org.komodo.rest.relational.connection.RestConnection;
+import org.komodo.spi.KException;
+
+/**
+ * An entity containing a connection and it's status.
+ */
+public final class RestConnectionSummary extends RestBasicEntity {
```
phantomjinx: A RestBasicEntity includes things like id and datapath so tends to represent an object from the komodo repo. To avoid that stuff with non-repo objects, extends AbstractKEntity instead (yes, naming conventions probably need to be updated).

server/komodo-rest/src/main/java/org/komodo/rest/relational/response/RestNamedVdbStatus.java
```diff
@@ -0,0 +1,87 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public RestMetadataVdbStatusVdb getVdbStatus() {
+        return this.status == null ? RestMetadataVdbStatusVdb.NO_VDB_STATUS : this.status;
+    }
+
+	@Override
+	public Object getXml() {
```
phantomjinx: If you extend AbstractKEntity rather than implement KRestEntity then this method is already done for you.

mdrillin:
We will go ahead and merge - and make note of the requested changes for the next PR.  We will have a related follow on PR soon...

PR #45, opened 2018-04-20
TEIIDTOOLS-312 Updates for MongoDB support

mdrillin:
Corrects the mongo property names.  Adds class-name to the properties.

### DIFF
--- a/server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/MongoDBDefinition.java
+++ b/server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/MongoDBDefinition.java
@@ -70,23 +70,28 @@ public class MongoDBDefinition extends DataSourceDefinition {
         {
               "DATABASE_SERVICE_NAME":"mongodb",
               "MEMORY_LIMIT":"512Mi",
+              "MONGODB_ADMIN_PASSWORD" "pwd",
               "MONGODB_DATABASE":"sampledb",
+              "MONGODB_PASSWORD": "pass",
+              "MONGODB_USER": "user",
               "MONGODB_VERSION":"3.2",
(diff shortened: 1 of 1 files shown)


server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/MongoDBDefinition.java
```diff
@@ -70,23 +70,28 @@ public Properties getDataSourceProperties(DefaultServiceCatalogDataSource source
...
+              uri
+              username
         }
         */
         Properties props = new Properties();
         props.setProperty("remoteServerList", source.getProperty("DATABASE_SERVICE_NAME")+":27017");
-        props.setProperty("username", source.getProperty("database-user"));
-        props.setProperty("password", source.getProperty("database-password"));
-        props.setProperty("database", source.getProperty("database-name"));
+        props.setProperty("username", source.getProperty("username"));
```
phantomjinx: Would it be useful to make these strings into constants at least so they only have to be modified in once?

rareddy: IMO, no need. I typically do the constants when I am sharing them in more than a single place. There is no guarantee these database properties are consistent or going to be consistent. They need to be treated as such.

phantomjinx: @rareddy 
Fine. No worries, just a passing thought.

mdrillin: I'll just leave as-is and merge it.  Can change later if needed

PR #49, opened 2018-05-15
TEIIDTOOLS-419: OData Test Editor

phantomjinx:


server/komodo-rest/src/main/java/org/komodo/servicecatalog/TeiidOpenShiftClient.java
```diff
@@ -632,213 +666,258 @@ private DeploymentCondition getDeploymentConfigStatus(DeploymentConfig dc) {
...
+        // if all threads are currently undergoing tasks. This ensures
+        // that the monitor thread will not try to do anything more with it.
+        //
+        work.setStatus(Status.CONFIGURING);
 
-        Vdb vdb = publishConfig.vdb;
-        String vdbName = vdb.getVdbName(uow);
-        try {
-            debug("Publishing (" + vdbName + ") - Start publishing of virtualization: " + vdbName);
+        threadService.execute(new Runnable() {
```
rareddy: Why another thread for the build? Was there a reason?

server/komodo-rest/src/main/java/org/komodo/servicecatalog/TeiidOpenShiftClient.java
```diff
@@ -1021,27 +1118,74 @@ public BuildStatus deleteVirtualization(String vdbName) {
...
             client.imageStreams().inNamespace(namespace).withName(vdbName).delete();
         } finally {
             kubernetesClient.close();
         }
-        runningBuild.statusMessage = "deleted";
+        runningBuild.setStatusMessage("deleted");
         return runningBuild;
     }
 
+    private RouteStatus getRoute(String vdbName, ProtocolType protocolType) {
```
rareddy: Maybe it would have been better to just return the coordinates of service or route and let the consuming utility take care of getting to the actual hosts and ports. But I guess that's OK

server/komodo-rest/src/main/java/org/komodo/servicecatalog/TeiidOpenShiftClient.java
phantomjinx: Configure takes a while. So as it is now initiated by the monitor thread, it would hold up the monitor thread from monitoring all the other builds. With 4 threads in the pool, up to 3 configures can be done in parallel along with the monitor thread monitoring. Thus, avoiding potential bottleneck.

rareddy: At a given request, there will be at most one configure thread + monitoring thread + calling thread in play. Previously I had the calling thread do the configure, now there will be one extra, which IMO not really needed because the aim is to free the calling thread and do everything in the secondary thread thus being ASYNC. Monitoring thread is also ephemeral once the build task is done it is terminated.

If a different publish request comes in the process that will get its own executor service and it's own Threads, as this is JEE servlet process is request/response model. The flaw as I can see even from the original design is there will be verbose monitoring threads if there are concurrent publish jobs running, we could tighten to only look at the work they are initiated with, not every work in the queue.

The status request should look at the shared status object, and return the result in the same calling thread, which it does.

phantomjinx: Calling thread has 30 second timeout being an http request so has to start things off then return.

Monitoring thread is ephemeral but there will always be 1 as long as there is work on the workQueue.

So there is not 1 extra since there remain 2 threads to do the work but both are not constrained by any timeouts and neither is constrained by the other in that the monitoring thread is free to continue monitoring other tasks as well as the configuring the other thread is performing.

None of the threads yield to each other so if there were multiple executor services, sooner or later the threads would get bottled-up waiting for a cpu slot. So better to limit the executor services to 1 and the number of threads to a sensible number. The executor service takes care of queueing tasks internally if a thread is not currently available.

The parallel instantiation of multiple monitor threads is an issue and better then to take this out of the executor service and keep as a single running background thread (sleeps if there is nothing in the workqueue).

PR #51, opened 2018-06-01
TEIIDTOOLS-420: Provides API for logging in publishing operations

phantomjinx:
* Adds REST API method to KomodoMetadataService for fetching the publishing
  logs for an individual virtualization being published.

* TeiidOpenShiftClient
 * Uses the info, error & debug functions to add logging to the main log
   file and to an additional file for the virt being published. This log
   can then be preserved by OS' storage.
 * The log file is available to be read by the REST API and returned as
   content for display in the UI.

### DIFF
--- a/server/komodo-rest/src/main/java/org/komodo/rest/KomodoRestV1Application.java
+++ b/server/komodo-rest/src/main/java/org/komodo/rest/KomodoRestV1Application.java
@@ -627,6 +627,11 @@ public class KomodoRestV1Application extends Application implements SystemConsta
          * Publish VDB
          */
         String PUBLISH = "publish"; //$NON-NLS-1$
+
+        /**
+         * Publish VDB Logs
+         */
+        String PUBLISH_LOGS = "publishLogs"; //$NON-NLS-1$
     }
 
     private static final int TIMEOUT = 1;
--- a/server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
+++ b/server/komodo-rest/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
@@ -2546,6 +2546,46 @@ public class KomodoMetadataService extends KomodoService {
         }
     }
 
+    @GET
+    @Path(V1Constants.PUBLISH_LOGS + StringConstants.FORWARD_SLASH + V1Constants.VDB_PLACEHOLDER)
+    @Produces( MediaType.APPLICATION_JSON )
+    @ApiOperation(value = "Find Publish Logs of Virtualization by VDB name", response = KomodoStatusObject.class)
+    @ApiResponses(value = {
+        @ApiResponse(code = 404, message = "No VDB could be found with name"),
+        @ApiResponse(code = 406, message = "Only JSON returned by this operation"),
+        @ApiResponse(code = 403, message = "An error has occurred.")
+    })
+    public Response getVirtualizationLogs(final @Context HttpHeaders headers, final @Context UriInfo uriInfo,
+            @ApiParam(value = "Name of the VDB", required = true) final @PathParam("vdbName") String vdbName)
+            throws KomodoRestException {
+
+        SecurityPrincipal principal = checkSecurityContext(headers);
+        if (principal.hasErrorResponse())
+            return principal.getErrorResponse();
+
+        KomodoStatusObject status = new KomodoStatusObject("Logs for " + vdbName);
+
+        List<MediaType> mediaTypes = headers.getAcceptableMediaTypes();
+        UnitOfWork uow = null;
+        try {
+            uow = createTransaction(principal, "publish", true); //$NON-NLS-1$
+            String log = this.openshiftClient.getVirtualizationLog(vdbName);
+            status.addAttribute("log", log);
+
+            return commit(uow, mediaTypes, status);
+        } catch (CallbackTimeoutException ex) {
+            return createTimeoutResponse(mediaTypes);
+        } catch (Throwable e) {
+            if ((uow != null) && (uow.getState() != State.ROLLED_BACK)) {
+                uow.rollback();
+            }
+            if (e instanceof KomodoRestException) {
+                throw (KomodoRestException) e;
+            }
+            return createErrorResponseWithForbidden(mediaTypes, e, RelationalMessages.Error.PUBLISH_ERROR);
+        }
+    }
+
     @DELETE
     @Path(V1Constants.PUBLISH + StringConstants.FORWARD_SLASH + V1Constants.VDB_PLACEHOLDER)
     @Produces( MediaType.APPLICATION_JSON )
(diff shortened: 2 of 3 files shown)


rareddy:
What is the reason for a fine-grained rest call and file-based management of log of deployment that is separate from other logging techniques? You could have folded the same info into existing publish status call and captured deployment log in that response and make it in single rest call from the client side. Seems like a verbose API to me.

phantomjinx:
@rareddy 

> What is the reason for a fine-grained rest call and file-based management of log of deployment that is separate from other logging techniques?

1. It is not separate from 'other logging techniques' in that all log entries are sent to the main KLog vdb-builder.log file as they always have been ever since the genesis of komodo. In addition to that logging, separate files for each of the published vdbs is generated (and recreated upon each new publishing attempt) so as to make it easy to return the log rather than having to filter KLog. Originally, these entries were kept in a hashmap and only available via the REST API. However, they are lost if the pod restarts or a new pod is deployed hence the file-based approach backed by persistent storage. (I shall create a JIRA to add that persistent storage to the das template).

2. At some point, I forsee KLog being retrofitted to the OS logging to allow for a dashboard/telemetry of how komodo is doing. Even allow DEBUG to be turned on/off (very chatty) if something needs to be looked at more closely. However, it would appear we are not at that point yet.

> You could have folded the same info into existing publish status call and captured deployment log in that response and make it in single rest call from the client side. Seems like a verbose API to me.

The virtue of a lean API lessens when the app performs poorly. By the time a published vdb is complete, the log is ~1700 characters. The publish status call is being made every 15 second. The log is only meaningful when displayed in the UI, ie. when the use opens the slide-in. Therefore, it makes no sense to keep delivering that same payload for all published vdbs that frequently. Thus, the verbose API may offend but it is more sensible to ensure the response is only received on an as-needed basis.

rareddy:
>Thus, the verbose API may offend but it is more sensible to ensure the response is only received on >an as-needed basis.
The question is about fine-grained API, If the message size is an issue, a flag can be added to the publish call to return the log or not, depending upon the log console visibility. That will keep related concerns in a single call. 

>It is not separate from 'other logging techniques' in that all log entries are sent to the main KLog vdb->builder.log file as they always have been ever since the genesis of komodo.
I do not know how KLog is different from Log4J like framework, but it should be that IMO, anything more is just more management on us.

> Originally, these entries were kept in a hashmap and only available via the REST API. 
I am not advocating to keep the whole log in Hashmap, but if you just want to keep the ephemeral publish status in memory that may be fine, especially if you use WeakhashMap where memory can be reclaimed if needed. On flip side it will get u out of the management of files.

>However, they are lost if the pod restarts or a new pod is deployed hence the file-based approach >backed by persistent storage. (I shall create a JIRA to add that persistent storage to the das template).
I do not think we need that, we are going to outsource the logging like what Steve is looking into for Open Tracing. So, individual applications should not be dealing with persistent storage, rolling the logs and updates etc. I see that is what you are saying in 2.

phantomjinx:
> The question is about fine-grained API, ...

Yes, and I have answered it. Your suggestion for conflating 2 API methods and separating them with a flag is useful but can ultimately result in an API lacking in clarity. However, if you wish to pursue this avenue then by all means file a JIRA and we can look at your contribution.

> I do not know how KLog is different from Log4J like framework, but it should be that IMO, anything more is just more management on us.

Once again your opinion is noted. FYI, KLog IS USING Log4J. KLog was actually ported over from Modeshape, at Randall's suggestion, as a means of maintaining a single logging API that sits on top of a number of different logging implementations (Log4J, slf4j and java.util.logging).

> I am not advocating to keep the whole log in Hashmap,

Good. Neither was I.

> but if you just want to keep the ephemeral publish status in memory that may be fine, especially if you use WeakhashMap where memory can be reclaimed if needed. On flip side it will get u out of the management of files.

Uh .... no. When vdb-builder has a restart I would still like to retain my publishing log, thanks, rather than have to guess when it was published and whether it went ok.

> On flip side it will get u out of the management of files.

Management of files is not onerous and ensures there is a written record of why things are going on.

> I do not think we need that, we are going to outsource the logging like what Steve is looking into for Open Tracing. So, individual applications should not be dealing with persistent storage, rolling the logs and updates etc. I see that is what you are saying in 2.

Really? Steve's adventures with Open Tracing have not yet born fruit have they? Now if you don't want the to retain logs (even in their current guise) in the das template then fine, I'll skip the JIRA then. Your call.

shawkins:
> Uh .... no. When vdb-builder has a restart I would still like to retain my publishing log, thanks, rather than have to guess when it was published and whether it went ok.

Any of the state management related to publishing needs to be persistent.  If you consider the log part of that state, then yes it would be persistent as well.

An important consideration here is to look at how this was done with Syndesis.  They have a fairly complex layer over standard pod logging - https://github.com/syndesisio/syndesis/tree/master/app/server/logging/jsondb - that scrapes pod logs to add entries to their postgres db instance (with retention schedules) and fronts it with rest service to look up events by integration/pod/etc.  This expects relevant events to already be logged in a json format with particular fields set so that it can be determined what the log relates to.  This is like a highly targeted logging stack, which would still play nice with the general aggregate logging.  If at all possible we should have similar approaches if not share their code.

> Really? Steve's adventures with Open Tracing have not yet born fruit have they? Now if you don't want the to retain logs (even in their current guise) in the das template then fine, I'll skip the JIRA then. Your call.

As much as possible we should stay out of the log infrastructure business.  

This is a special case in that the log can be considered part of the publishing state.

OpenTracing is for tracking things from a high level perspective - so that you have a call trace graph of the operation across a distributed environment that you can look at just like you would if it were from a single system.  We could have image publishing instrumented that way, but it would mostly be of interest to us and not end users / developers.

For simple logging statements just allowing those to go to the console is sufficient.  I'm evaluating how much it helps to have them in a logstash format - but in general we want to weed out any of the custom log handling we're doing (multiple log4j appenders, separate files, etc.).

phantomjinx:
@shawkins 
> that scrapes pod logs to add entries to their postgres db instance

I considered this since we have a postgres db sitting around handling the komodo repository persistence. Shouldn't be too hard to construct a schema for the logs. At the time, given the infancy of this part of the functionality I felt it was overkill. In the end, simply preserving the publishing logs was the requirement, so a written file to a persistent filesystem mounted on /opt/app-root/src/logs does the job.

The current state of play (away from the publishing special case) in komodo is that KLog is the logging-king in that all errors, info, debugging are done through it. As stated, it is able to use three different logging impls depending on what it is installed with (currently Log4j which logs to the console and to the log/vdb-builder.log file) so it will be very easy to generate an impl without needing to touch most of the codebase.

If you would like to propose a JIRA with requirements / proposed implementation framework etc... then happy to look at it.

rareddy:
> Uh .... no. When vdb-builder has a restart I would still like to retain my publishing log, thanks, rather than have to guess when it was published and whether it went ok.

Logging entries are not state. Yes, publishing status needs to be persisted or in our case, I believe most of it can be queried from cluster to establish the state. Logging entries corresponding publishing progress is up to us, whether that is critical information to keep and go through complex interactions to persist it, or you think it is a by-product and can be dispensable. The real log whether it is KLog, Postgres, Log4J, logstash will still have the all entries from before pod restart.

phantomjinx:
> go through complex interactions to persist it
Honestly, the interactions were not complex. In fact, they were rather straightforward. If you would like to peruse the PR to check ...

Like I said, your opinion is noted.

PR #52, opened 2018-06-04
TEIIDTOOLS-313: Adding salesforce connection capabilities

rareddy:
* fixed some resource adapter creation properties
* Added example to teiid-throntail repo to showcase usage of salesforce connection standalone

### DIFF
--- a/server/komodo-rest/src/main/java/org/komodo/servicecatalog/DataSourceDefinition.java
+++ b/server/komodo-rest/src/main/java/org/komodo/servicecatalog/DataSourceDefinition.java
@@ -93,7 +93,13 @@ public abstract class DataSourceDefinition {
     protected void ds(Properties props, DefaultServiceCatalogDataSource scd, String key, String value) {
         if (scd.getDefinition().isResouceAdapter()) {
             props.setProperty(
-                    "swarm.resource-adapter.resource-adapters." + scd.getName() + ".connection-definitions." + key,
+                "swarm.resource-adapters.resource-adapters." 
+                        + scd.getName() 
+                        + ".connection-definitions." 
+                        + scd.getName() 
+                        + ".config-properties."
+                        + key
+                        + ".value",
                     value);
         } else {
             props.setProperty(
@@ -102,6 +108,21 @@ public abstract class DataSourceDefinition {
         }
     }
 
+    protected Properties setupResourceAdapter(String dsName, String moduleName, String className, String jndiName) {
+        Properties props = new Properties();
+        // consult Teiid documents for all the properties; Then map to properties from
+        // OpenShift Service
+        String raPrefix = "swarm.resource-adapters.resource-adapters." + dsName + ".";
+        String cdPrefix = raPrefix + "connection-definitions." + dsName + ".";
+        props.setProperty(raPrefix + "module", moduleName);
+        props.setProperty(raPrefix + "transaction-support", "NoTransaction");
+        props.setProperty(cdPrefix + "class-name", className);
+        props.setProperty(cdPrefix + "jndi-name", jndiName);
+        props.setProperty(cdPrefix + "enabled", "true");
+        props.setProperty(cdPrefix + "use-java-context", "true");
+        return props;
+    }
+    
     /**
      * If source is provisioned through a service catalog this routine returns true. For sources like file, webserive
      * this will return false.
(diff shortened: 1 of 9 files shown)


server/komodo-rest/src/main/java/org/komodo/servicecatalog/DataSourceDefinition.java
```diff
@@ -93,7 +93,13 @@ public boolean isResouceAdapter() {
...
         if (scd.getDefinition().isResouceAdapter()) {
             props.setProperty(
-                    "swarm.resource-adapter.resource-adapters." + scd.getName() + ".connection-definitions." + key,
+                "swarm.resource-adapters.resource-adapters." 
+                        + scd.getName() 
+                        + ".connection-definitions." 
+                        + scd.getName() 
+                        + ".config-properties."
+                        + key
+                        + ".value",
```
phantomjinx: Can you provide a quick explanation on
a) why this has changed;
b) why the much longer name
c) how this would impact both salesforce and other resource adapters
d) is this a breaking change?

server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/AmazonS3Definition.java
```diff
@@ -29,7 +30,7 @@
 
     @Override
     public String getType() {
-        return "amazon-s3";
+        return "webservice";
```
phantomjinx: Why is this now more generic given its in the named class, please?

server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/MongoDBDefinition.java
```diff
@@ -98,23 +98,17 @@ public Properties getDataSourceProperties(DefaultServiceCatalogDataSource source
...
+                "org.teiid.resource.adapter.mongodb.MongoDBManagedConnectionFactory", jndiName);
         ds(props, scd, "SecurityType", "SCRAM_SHA_1");
         ds(props, scd, "Ssl", "false");
         ds(props, scd, "AuthDatabase", scd.canonicalEnvKey("database_name"));
         ds(props, scd, "RemoteServerList", scd.canonicalEnvKey("DATABASE_SERVICE_NAME")+":27017");
         ds(props, scd, "Database", scd.canonicalEnvKey("database_name"));
-        ds(props, scd, "Username", scd.canonicalEnvKey("database-user"));
-        ds(props, scd, "Password", scd.canonicalEnvKey("database-password"));
+        ds(props, scd, "Username", scd.canonicalEnvKey("username"));
+        ds(props, scd, "Password", scd.canonicalEnvKey("password"));
```
phantomjinx: Are these not in danger of falling foul of one mongodb implementation despite working with another? Might it be safer to simply apply both sets thereby covering all bases since its difficult to know which are preferred?

server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/ODataV4Definition.java
```diff
@@ -29,7 +30,7 @@
 
     @Override
     public String getType() {
-        return "odata4";
+        return "webservice";
```
phantomjinx: See Amazon class comment. Is this getType() more generic now?

server/komodo-rest/src/main/java/org/komodo/servicecatalog/DataSourceDefinition.java
rareddy: No, it is not breaking change, it was never correct to begin with. The properties were designed before publish operation. See example here https://github.com/rareddy/wildfly-swarm-teiid-examples for properties required to create a RA.

server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/AmazonS3Definition.java
rareddy: This represents the resource-adapter internally.

server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/MongoDBDefinition.java
rareddy: no, they were placeholders before, same with other sources even now. I do not think Mark ever did the publish operation on MongoDB. Thus did not get to this initially. (Actually, I did not do in OpenShift either because of state of the Beetle application in repo currently)

server/komodo-rest/src/main/java/org/komodo/servicecatalog/datasources/ODataV4Definition.java
rareddy: same as above.

PR #84, opened 2018-11-08
Upgradeto6

tejones:


rareddy:
Yeah! does this mean the plugin component is working?

tejones:
Not quite @rareddy.. we are close though. This commit was just some path changes.

PR #105, opened 2018-11-29
TEIIDTOOLS-466: Adding OpenTracing support

rareddy:


shawkins:
LGTM - the last step is to have the agent running with the application - described as a sidecar https://github.com/jaegertracing/jaeger-openshift 

istio may also provide that automagically.

PR #117, opened 2018-12-17
ng tests should be in a profile so they can be disabled in a headless build environment

cunningt:
Move ng tests into a profile so they can be disabled in a headless build environment.


cunningt:
Also, aligned spring/spring-boot with Fuse versions, as well as jackson-databind

rareddy:
Thanks @cunningt

PR #137, opened 2019-05-30
TTOOLS-630 Added ddl parsing to view definition validation

blafond:


server/komodo-server/src/main/java/org/komodo/rest/service/KomodoUtilService.java
```diff
@@ -121,6 +121,11 @@
     public static final String USER_NAME = "User Name"; //$NON-NLS-1$
 
     public static final String WORKSPACE = "Workspace"; //$NON-NLS-1$
+    
+    /**
+     * Time to wait after deploying/undeploying an artifact from the metadata instance
+     */
+    private final static int DEPLOYMENT_WAIT_TIME = 10000;
```
mdrillin: you can remove - not used

server/komodo-server/src/main/java/org/komodo/rest/service/KomodoUtilService.java
```diff
@@ -1233,4 +1168,8 @@ public Response removeViewEditorState(final @Context HttpHeaders headers,
             return createErrorResponseWithForbidden(mediaTypes, e, RelationalMessages.Error.PROFILE_EDITOR_STATE_REMOVE_ERROR);
         }
     }
+
```
mdrillin: can remove - dont think this is used anymore

server/komodo-server/src/main/java/org/komodo/rest/service/KomodoUtilService.java
```diff
@@ -968,108 +972,39 @@ public Response validateViewDefinition(final @Context HttpHeaders headers,
...
-			} catch (Exception e1) {
-			}
-        }
-        
-        cleanupValidationFile(inputStream, tmpFile);
+        UnitOfWork uow = null;
         
-        // create response
         try {
+        	SynchronousCallback callback = new SynchronousCallback();
```
mdrillin: you dont need SynchronousCallback anymore - was using it before because there were 2 transactions.  Just do the commit like the other methods.  

return commit( uow, mediaTypes, viewDefnStatus );

also, rollbackOnly should be true on the createTransaction since you arent changing anything (think I had that wrong originally)

uow = createTransaction(principal, "Validate ViewDefinition", true );

server/komodo-server/src/main/java/org/komodo/rest/service/KomodoUtilService.java
blafond: done

server/komodo-server/src/main/java/org/komodo/rest/service/KomodoUtilService.java
blafond: done

server/komodo-server/src/main/java/org/komodo/rest/service/KomodoUtilService.java
blafond: done

PR #170, opened 2019-07-11
Refactor

shawkins:
This separates things more so that only core has direct modeshape/teiid-modeshape dependencies - including any lexicon references.

This probably has more logic changes than the other pull requests that focused more on removal, so it will need more testing.  It looks fine to me on an initial pass, but I'm not sure about some behavior - such as I can't seem to change view description via the edit box, rather I have to modify the annotation option in the sql.


rareddy:
looks good

PR #172, opened 2019-07-12
removing komodoobject from spi

shawkins:
With the exception of the initialization classes ApplicationProperties and PersistenceType, this removes the bulk of the jcr/komodoness (KomodoObject) out of the spi.  The UnitOfWork is still omni-present, but that is a minor issue really.  What is left is much tighter set of interfaces and utility classes, but it still is leaking the path abstraction WorkspaceManager.getKomodoWorkspaceAbsolutePath and Dataservice.getAbsolutePath.

Other than some cleanup, what should happen is that actual business logic (ServiceVdbGenerator and anything else) needs promoted out of core into the server.  Then core can be replaced with relative ease.

Or we could instead look to cleanup some of the server logic now that we have a constrained paradigm for dealing with the persistence layer.


rareddy:
What you ideas behind server logic? As I was saying with different vdb this is also key to reduceing complexity. That would be good.

shawkins:
I should add that this fixes an issue with the connection loading/active flag that was broken with the last pr.

> What you ideas behind server logic? As I was saying with different vdb this is also key to reduceing complexity. That would be good.

I was hoping that something would naturally emerge as the refactoring progressed, but this work isn't that high up.  The current state basically has 4 vdb methods (create, delete, find 1, find many), 4 similar dataservice methods, and 2 profile related - but offers no insight on the interplay between those.  

It's simple enough to take a pass through and remove parts of the the spi interface that are not used by the server, but that doesn't quite get at the spirit of different or differentiated vdbs (schema vdb, connection vdb, service vdb...).  

Assuming that we want to hide the vdb concept as much as possible even from the server layer, what's a good high level description of how all of the various vdbs are used?  It should be straight-forward from there to better isolate the vdb concept.

PR #179, opened 2019-07-19
consolidate rest services

mdrillin:
NOTE: has companion PR on ui.  Let me know before merging.
- Removed "getViewEditorState" from KomodoUtilService.  It was not used - and can use other method to get single state by passing in a name pattern.
- Eliminated "{syndesisSourceName}/schema" service method from KomodoMetadataService.  Modified /connection-schema service method to take a name pattern to provide equivalent functionality.

### DIFF
--- a/server/komodo-server/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
+++ b/server/komodo-server/src/main/java/org/komodo/rest/service/KomodoMetadataService.java
@@ -21,7 +21,6 @@ import java.net.URI;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


shawkins:
The only thing that concerns me is the notion of what exactly is the search pattern syntax.  The replacement of * with .* implies the client expects globs - however with that conversion any of the other regex metacharacters would be a problem.  Do we have any name validation logic that ensures dataservices can't be created with names that have such characters?

In general there's some cleanup that we'll need to do where names are substituted for patterns - but without considering metacharacters - or using globs some places and regex others.

mdrillin:
No, I don't believe there is any checking of the name characters currently.  So yes that is a deficiency.  The same problem exists on other methods that use the same approach.  Since this PR was just cleanup / removal of a couple methods, I'll just close it in lieu of future clean up.

PR #180, opened 2019-07-22
Removing the modeshape service vdb

shawkins:
This removes the modeshape storage of the service vdb - leaving only the schema vdbs.  For now the refresh views service method is not doing anything, rather the service vdb is built on demand with publishing from the view editor states.  This of course could be refined such that the refresh method creates a deployed version of a virtualization specific vdb which could be reused if it exists on a publishing call.


rareddy:
:+1:

PR #181, opened 2019-07-22
Add a teiid-springboot dependency

cunningt:


rareddy:
@cunningt Why is this dependency being added?

cunningt:
@rareddy version.springboot.teiid isn't getting aligned because there's no accompanying dependency associated with it.     If we could add a dependency here that uses that property, I think it would fix the issue in https://issues.jboss.org/browse/ENTESB-11073

rareddy:
ok, merging it.

PR #185, opened 2019-07-24
Completing the removal of modeshape vdb storage

shawkins:
This changes removes all vdb related logic from modeshape.  There is the storage of a the source schema as just a string instead.  If it exists it will be used as the basis for the source/connection vdb.

This also inlines the logic for the deletion of the schema into the connection synchronizer.  The only issue with this approach is that there's no explicit mechanism to force the refresh of source metadata (from a ui action or timer), but that's not something that we're doing yet.


shawkins:
I should add that this removes waits around deploy/undeploy because embedded operations are synchronous.

shawkins:
Updated the pr that calls the servicevdbgenerator on refresh as well.  That should evolve into a virtualization specific preview vdb, but requires some changes for that.  So for now we'll just generate the service vdb on refresh and publish.

PR #186, opened 2019-07-25
TEIIDTOOLS-712 removing the profile concept

shawkins:
and updating the swagger link to openapi


shawkins:
I went ahead and merged this as later today or Monday, we'll look toward full modeshape removal.

PR #188, opened 2019-07-27
Refactor

shawkins:


shawkins:
This completes the modeshape removal.

The spi module (could be renamed common) contains just three packages, komodo, komodo.datavirtualization, and komodo.utils.  The Dataservice interface has been renamed DataVirtualization.

The core module (could be renamed repository) contains just a single package komodo.repository implementing the entity model and workspacemanager, etc.  A case could be made for combining everything into a single module, but this seems fine for now.

ViewEditorState has been replaced by ViewDefinition and the state classes have been removed.

The schema.sql file has an initial guess at some string widths for names, descriptions, and even the ddl.  All of that can be tweaked or just set to the max value to emulate the modeshape behavior.

Everything seems to come up for me in openshift, but the ui is using the getDataservices method that takes a pattern, which is no longer allowed.

rareddy:
I think you should just merge and continue. May be difficult to review with large commit set.

PR #189, opened 2019-07-30
TEIIDTOOLS-718 directly wiring the connection synchronization logic

shawkins:


shawkins:
@rareddy see if this seems ok to you.  This removes the service layer from the syncronizer calls - which allows us to remove a couple more service methods.  The only real difference is the removal of the retry logic, that I don't think that is needed with this change.

rareddy:
Great, I see you converted to them to method calls from service calls, that make it no need to re-try logic.

PR #192, opened 2019-07-31
Refactor

shawkins:


shawkins:
This is a first step toward naming changes:

- source schema is tracked by syndesis id, so there's no chance that we'll use the wrong source metadata
- when we create a komodo representation of a given syndesis source we'll cleanse/munge the name into a unique komodo/teiid schema name

This means that most of the time the connection name will be the schema name that we use.  If need be the ui could display both the syndesis connection name and the teiid schema name on our connection cards.

shawkins:
I don't see that the ui is using RestSyndesisSourceStatus.schemaModelName - this commit does change that to schemaModelId.  It could be changed to just id as well.

PR #197, opened 2019-08-02
TEIIDTOOLS-723 removing explicit json parsing and a swagger workaround

shawkins:


shawkins:
This address the swagger model not showing properties - the problem is with the swagger / jackson logic.  If you use a JsonSerialize annotation, but don't specify the as you don't get any properties.  I'm not sure if that's a known issue, but it's definitely a bug as the property is not required.

PR #198, opened 2019-08-05
TEIIDTOOLS-723 TEIIDTOOLS-724

shawkins:
* removing komodorestexception
* removing more serializers
* updating schema naming to use the db
* merge the rest definition of ViewEditorState and ViewDefinition
* explicitly associating viewdefinitions to datavirtualizations - now the dataVirtualizationName is required
* ViewEditorState id is now the UUID assigned as creation time
* ViewEditorState viewName is replaced by ViewDefinition name
* the stash method takes only a single viewdefinition, not an array, and
returns the generated id in the response.

This is a db schema breaking change - you'll need to recreate your pg instance for this.


rareddy:
nice, lgtm

PR #202, opened 2019-08-12
Refactor

shawkins:


shawkins:
This addresses the all column convention by adding an "advanced" flag to the view definition.  When true (the default), the user is in an advanced editor mode - they can directly modify the ddl and the sql compositions, projected columns, etc. will not be used to construct the ddl.

Given that there is already a userDefined flag, that gives us 4 states (userDefined/advanced):

t/t - the user has directly modified the ddl
f/t - the user can directly modify the ddl, but it is still the generated default
t/f - ???
f/f - the user cannot directly modify the ddl, and the ddl is generated by the view editor state

In other words it appears that we only have 3 states.  @mdrillin @blafond - does this match your understanding?  And would you replace these booleans with a single enum?

mdrillin:
I don't understand how the 'advanced' flag relates to the 'allColumns' convention.  Also, what is the scenario where we would want to prevent the user from modifying the DDL?

shawkins:
> I don't understand how the 'advanced' flag relates to the 'allColumns' convention. 

After thinking about it more, we don't want to add a bunch of flags such as "allColumns" to the viewdefinition.  What we're really trying to determine is if the user is directly editing ddl (advanced) or if they are using some kind of guided editor which has state that includes projected columns, compositions, etc.

Currently there's only 2 states - either the user have modified the ddl, or it's been generated by the server.  

So what I'm getting at is that there will eventually be a third state - the user is using a guided editor.  If the user wants to edit the ddl generated by using sql compositions and the like, then more than likely all of that extra state will become invalid - without a lot of additional work that will allow us to see if the user sql matches up with something that can be represented by the editor.

If we don't want to tackle the state for that yet, we can just leave things with the userDefined boolean - and even remove the projected columns and sql compositions until we have a better handle on what the guided editor will be doing.

> Also, what is the scenario where we would want to prevent the user from modifying the DDL?

If you use the guided editor, directly editing the generated ddl will likely invalidate the additional editor state, so there would likely be some explicit toggle to switch to the advanced editor.

mdrillin:
My thinking on the userDefined flag was that once the user edits the ddl, then userDefined=true - and the guided editor controls would disable.  But yeah I can see where there would be a third state - if the edited DDL matched the current state of the editor controls.

So I was thinking there would not be an explicit 'switch' between guided and advanced.  The user would be warned when they changed the DDL that doing so would invalidate the editor controls.  But I'm open for suggestions

shawkins:
> So I was thinking there would not be an explicit 'switch' between guided and advanced. The user would be warned when they changed the DDL that doing so would invalidate the editor controls. But I'm open for suggestions

As long as there's some acknowledgement that the direct ddl editing can invalidate the guided editor.

To explain things a little differently, the old code when user defined was false the intention of when to use the projected columns (or even the compositions) was muddied by the all column convention.  An additional flag clarifies that.

If we don't want an additional flag, then we can just remove the projected columns and compositions and stick with the user defined flag.

PR #203, opened 2019-08-14
TEIIDTOOLS-739 TEIIDTOOLS-726

shawkins:
removing pagination
logging/util cleanup
using teiid names instead of source fqn paths


shawkins:
This has some minor stuff, such as removing the custom string formatting code that was part of the KLog and other util cleanups.

The big thrust is to change sourcePaths from "connection={conn/schema name}/{source fqn}" to "schema={schema name}/table={teiid table name}" - the rationale for that is explained on the issue.

Other changes related to the UI:

* RestViewSourceInfo / viewSourceInfo - this change is not fully realized.  It removes paths off of the response objects, and it does show removing the scoping to just a view.  I've been trying to discern from Barry what the scope of the auto-completion is, so I'm not sure he's on board with this change.  In short restricting to only the schema that are already referenced doesn't capture everything that is potentially legal to reference - currently it's everything exposed by the preview vdb.  Assuming there's agreement we can refine the method name and response object name from here.  Also I'd like for this metadata / auto-complete to be the basis for a generalized sql client / preview screen that allows for the user to execute any sql.

* RestSchemaNode - removed the path property and are now using the teiidName property to capture the teiid name - which is what needs to be specified in the import list.  If there's a need for the ui to have the fully constructed teiid fqn "schema={schema name}/table={teiid table name}", we can add that.

blafond:
Since the UI's TableColumns object is unique to the UI and the DdlEditor's code-complete feature, the ViewSourceInfo Response can really be of any form and content as long as we can access Table/Column names.

The sourcePath doesn't have to be returned, but I thought we'd need to maintain some "source schema" referenced in the Response for the UI to sync any user changes to the current ViewDefinition.

I'm on board with broadening the metadata that's exposed to the UI

shawkins:
Updated get viewSourceInfo to getRuntimeMetadata - it will take a virtualization name to eventually return the metadata specific to a virtualization.

I'm using the term runtime to specifically denote that only "complete" views will be included.  Eventually if the editor state can fully capture what columns are supposed to be on a view, we could include incomplete metadata as well.  Note that everything gets more difficult when we allow views to reference other views because the notion of completeness becomes transitive.

PR #205, opened 2019-08-16
Refactor

shawkins:


shawkins:
This adds a virtualization specific preview vdb.  Each complete view will be available.  The preview query logic has changed to expect a virtualization name, rather than a vdb name.  The ui logic for a preview query should be changed to just issue "select * from view" - no ddl parsing nor usage of the source paths is needed.

This adds a dirty flag to the datavirtualization to let us know when to refresh it on preview.

Once we address allowing views to reference other views, this will get more complicated.

shawkins:
I should clarify that it needs to be a fully qualified preview query:

select * from views.viewname

Because the source tables are not hidden in the dv specific vdb, so they could have an ambigious name.

Note that the views schema name convention is worth discussing.  It will prevent any layering of data vdbs - as there would be schema collision on import.  If we want to eventually allow for that, then we may want to choose a different convention, such as the data virtualization name - but of course with restrictions to make it a valid schema name.

rareddy:
> select * from views.viewname
Isn't this single "virtual model" based view?

>Because the source tables are not hidden in the dv specific vdb, so they could have an ambiguous >name.
From the UI there is no direct access to the user, so why not classify them as hidden?

shawkins:
> Isn't this single "virtual model" based view?

I'm not sure what you are asking.

> From the UI there is no direct access to the user, so why not classify them as hidden?

I'd like for the dv specific preview vdb to become the basis for the a generic sql client, such that you can execute any valid sql in the context of the dv - which means that the source tables cannot be hidden.

rareddy:
I am saying that we agreed there is going to be a single virtual model in the vdb, so the resolution always needs to be at the virtual model level.

>I'd like for the dv specific preview vdb to become the basis for the a generic sql client, such that you >can execute any valid sql in the context of the dv - which means that the source tables cannot be >hidden.
On this, I would differ. The generic SQL preview should be open to the views exposed through vdbs. We do not have to stop the user from executing the source level queries if they know the full path. as such if we do it right the SQL editor can support auto-compete for these tables.

shawkins:
> I am saying that we agreed there is going to be a single virtual model in the vdb, so the resolution always needs to be at the virtual model level.

I'm not entirely sure, but you're probably just communicating the same thing as what I'm expressing - when the ui is responsible for issuing a preview query for a given view, then it should be "select * from views.view".  If the sources were hidden, then it doesn't need to be qualified, but as I'm indicating there's a reason for why you would want a general dv specific preview vdb.

> On this, I would differ. The generic SQL preview should be open to the views exposed through vdbs. We do not have to stop the user from executing the source level queries if they know the full path. as such if we do it right the SQL editor can support auto-compete for these tables.

I'm suggesting a mode for a sql client is a workbench for playing around with sql/results until you're ready to create a view from it.  In that mode you should be able to reference other views and sources.

There's another preview mode of just the "views" - which is currently restricted to a drop down list.

shawkins:
This discussion made me double check, and sure enough the source models were not hidden in the service vdb, so that has been updated as well.

rareddy:
So you only suggesting that when drop down queries coming in always qualify them with `view.` which is static. If we populate them with qualified name to begin with that issue should be solved.

I do agree the ad-hoc mode is good, that is what we wanted to support to begin with, but I guess we choose current one for expediency.

shawkins:
> If we populate them with qualified name to begin with that issue should be solved.

The user won't see that.  The ui currently has logic that either tries to parse the view ddl, or tries to construct a select query from the sourcePaths.  I'm saying that the ui should replace all that logic with just "select * from views.view"

rareddy:
yeah, from what I read from your previous comments that need to be taken out anyway, that will not work.

PR #207, opened 2019-08-16
Correcting the dirty flag migration, spring boot 2, and refinements to the pom

shawkins:


shawkins:
This incorporates both the outstanding pull requests.

shawkins:
Made more changes that are needed once we launch in container - in particular flyway needs to do a migration.  Rather than require an intemediate app installation, I've added a bean as described in the spring 2 docs.

PR #209, opened 2019-08-16
TEIIDTOOLS-738: Updating to Spring Boot 2.1.3 and removing the Jersey…

rareddy:
… web engine based JaxRS based web API


rareddy:
Before @shawkins does another big one, please take look and merge. I did verify with UI. BTW UI response seems much snappier now :)

shawkins:
Thanks Ramesh.  The story of the the incredible shrinking komodo continues :)

A couple of small things / questions:

* what does the default exception handling look like now that the exception mapper is gone?  It looks like ErrorResponse is no longer used, so presumably there's a default json error mapping by spring, correct?
* the service method javadocs will need updated to remove the headers/uriinfo
* several of the protected methods, such as getSchema, were changed to getSchemaService.  Those methods were mostly to bypass any of the rest stuff, so now it would be better if they were merged with the calling method.  The only other change to support that is to the transactional logic to use TRANSACTION_REQUIRED for the propagation - as there's automatically a test transaction created and we don't want a nest transaction.
* It would be nice to have the MediaType constants in things like the RequestMapping

rareddy:
Thanks Steve.

* Added test case for Exception.
* corrected the javadoc
* added media type
* I did not quite understand your transaction comment. So I left as is for now.

Also I feel we need to ditch the all String constants on the `@RequestMapping` I can never tell from the long concatenated strings what they are, can we clean up as par of rest cleanup, please?

shawkins:
I updated my other pr to include this one with some changes.  The only real issue were a couple of places where changes from other commits had been lost.  See https://github.com/teiid/teiid-syndesis/pull/207/commits/a3e41d980c782177a40423c761aa80c80da791b0

rareddy:
Ok closing this PR then.

PR #210, opened 2019-08-20
TEIIDTOOLS-738: Errors with CORS, setting of SB2 specific servlet con…

rareddy:
…text, and marking the POST,PUT requests with BODY payloads


blafond:
Pulled locally land was successful in deploying a syndesis-dv pod :)

rareddy:
super! There are a couple of issues on UI side and backend side which do not involve with SB2 migration, which are related to view creation. I will see if I can submit PR for that.

PR #211, opened 2019-08-21
updating the merged transactional logic

shawkins:


shawkins:
my initial merge of the transaction code did not properly account for read-only transaction rollbacks.

PR #212, opened 2019-08-21
TEIIDTOOLS-751: 

rareddy:
1) Correcting view editor namevalidation endpoint annotations
2) View deletion always returning 404, because of the duplicate deletes, which is corrected
2) Updating MySQL version as current version does not work with Spring Boot 2 or Java11
3) when undeploying the VDB, do not use "-vdb.xml" name, use vdb name
4) Lots of checkstyle issues, not sure why did not pop up before


shawkins:
LGTM

PR #215, opened 2019-08-28
TEIIDTOOLS-741 adding better concurrency control

shawkins:


shawkins:
This adds a base class to our db entities that includes optimistic locking, database enforcement of case-insensitive names, and created/updated times.  Most concurrency exceptions at this point are not caught - if two users try updating the same view as the same time, one of them will get an exception.  We can clean that up later if needed.

Some refactoring was done to move updates to the runtime state until after the related database commit - and where relevant using a lock to ensure consistency.  The only thing that has any real complexity is the refreshing of source schema.  The code was updated to allow for a ui directed refresh of the metadata, but as a blocking operation.  Thinking about it more we probably want that functionality as non-blocking (for example adding a task to the connection synchronizer queue).

The current code mostly allows for multiple server pods.  However there is no formal mechanism for synchronizing schema across the runtime state of the pods.  I can add a subtask related to that.

shawkins:
Actually I think I'll just merge in the view layering changes with this.

PR #216, opened 2019-08-30
adding better concurrency control and view layering

shawkins:
This adds view layering and refines the early concurrency commit - view editing optimistic locking requires that the version specified in the json is provided back by the ui.  If not, it will always replace the latest.  The rest of the concurrency issues / remaining work are discussed on the jira.

The biggest missing piece here is a mechanism of providing validation / parsing errors - with some minor modification to the Teiid metadata validation, we can associate the error not just to the model, but to an object as well, so that we could provide a list of all validation/parse errors pretty quickly for a data virtualization from looking at the dv specific preview vdb.

getRuntimeMetadata was updated to provide all metadata - although it will provide the view itself in that.  Another missing bit is that the getSchema / getConnectionSchema are specific to source/connection metadata - they either need to be updated or new methods for the view level metadata as well.

Many operations rely on having an always available / up-to-date dv preview vdb.  This is costly in terms of rebuilding them, and in terms of memory footprint.  Effectively all metadata will be in memory twice - once on the vdbmetadata, and again on the runtime objects.  A mitigation strategy could be to only allow the vdbs to live for a short time - which would also address any concern about cleanup after the removal of a data virtualization.  We're implicitly assuming that there won't be a similar issue with the source metadata, such that we can leave the base preview vdb available.  Alternative designs include a querymetadatainterface wired to the database.  However that will work for validation, but not for preview.  As with designer there's too much logic that becomes dependent upon the metadata objects and the environment, such that a full vdb is expected.  So if we are going to have a tightly coupled preview feature, there's probably little benefit in going down the path of having a db backed querymetadatainterface.

This design allows for arbitrary view layering, see the IntegrationTest for example.  This of course is not the only design.  An alternative would be to allow for only a single level of layering, but allow the user to have virtualizations reference one another.  That introduces a different set of complications - as mentioned on a jira is the choice to have a single virtual model name, which prevents us from from using the current import logic which requires distinct schema names.  It would also probably require different management from connections - otherwise you'd force the user to re-import from each referenced virtualization.


rareddy:
I tried to look through the code, but with amount of code changes, I can't really say I grasp all the changes. I am sure they are fine.

For validation and parsing errors and context-sensitive help, there was a big push towards [LSP](https://microsoft.github.io/language-server-protocol/) at F2F we should investigate and invest in this.

for multi-vdb layering, we could consider the "view" name is same as the "vdb" name then we will solve the collision of schema names.

shawkins:
> For validation and parsing errors and context-sensitive help, there was a big push towards LSP at F2F we should investigate and invest in this.

I'm good with that.

> for multi-vdb layering, we could consider the "view" name is same as the "vdb" name then we will solve the collision of schema names.

Let's come to a consensus on https://issues.jboss.org/browse/TEIIDTOOLS-753 - if we make the dv name the view schema name, then we'll just need to add some additional validation or introduce logic that will convert it into a valid non-reserved schema name (just as we are doing with the syndesis source schema names).

rareddy:
> So if we are going to have a tightly coupled preview feature, there's probably little benefit in going down the path of having a db backed querymetadatainterface.

As I see this will relieve us from constant redeploying of the VDBs and minimize on size of the runtime objects. This is similar to non-deployment based model. I did not understand why you said this will not work for preview?

shawkins:
> I did not understand why you said this will not work for preview?

Because it hasn't worked that way for years.  The querymetadatainterface works for resolving/validation.  Beyond that there's a lot of execution logic that is dependent upon the actual metadata objects, not just the the querymetadatainterface.  All of that would need to be redone/revalidated against a dynamic metadata source providing actual metadata objects.  Think of all the logic in the translators directly referencing Table, etc.  Until we bite off that work, we won't have a fully embedded execution path that is not dependent upon vdb deployment.

rareddy:
In translators, we use RuntimeMetadata, I was under the assumption that is backed by QMI, if that is not the case yes, that is an issue.

shawkins:
> In translators, we use RuntimeMetadata, I was under the assumption that is backed by QMI, if that is not the case yes, that is an issue.

Look at RuntimeMetadataImpl it casts things coming out of the qmi to actual metadata objects, rather than further usage of qmi.

I'm not saying it's impossible for things to work differently, just not possible currently.  It requires implementing a qmi backed by the database that also converts to actual metadata objects.  Of course queries involving system tables get hairy - generally that would mean that the vdb will need to be fully realized.  But of course we can also begin to look again at alternative sys logic (the old version was based upon the index files - low memory usage, but bad performance).

mdrillin:
So as far as UI compatibility with this, in testing - it looks like I will just need to change the preview queries - (WAS)
{"limit":15,"offset":0,"query":"SELECT * FROM views.winelist","target":"testVirt2"}

In the above, remove 'views.' - so would be SELECT * FROM winelist  ?

shawkins:
> The above query would remove 'views.' - so would be SELECT * FROM winelist ?

With these changes, that should still be views.winelist.  

https://issues.jboss.org/browse/TEIIDTOOLS-753 is discussing the views schema name convention.

mdrillin:
> With these changes, that should still be views.winelist.

There must be a different issue at play then, because the views.winelist query fails.

shawkins:
Hopefully an appropriate error is logged?

rareddy:
> The above query would remove 'views.' - so would be SELECT * FROM winelist ?
IMO, this is the correct way to issue the query, it should not assume any model/schema names. as per the conflict of the names with source models, my position is to have them hidden.

shawkins:
> IMO, this is the correct way to issue the query, it should not assume any model/schema names. as per the conflict of the names with source models, my position is to have them hidden.

If you recall this was a topic on another issue.  Short answer is that yes for now the qualification is needed.  If we don't want the ui to have that knowledge then it will need to have another query method that just takes the view name.  

The longer answer is that there's no reason to restrict the dv specific preview vdb in the way that we are - there should be a sql client preview capability where the user can execute whatever sql they want and then use that as the basis for a view.  That requires a preview vdb where the source schema are not hidden.  Note that once we save something as a view it becomes much more heavy-weight.  The more we can do with just arbitrary sql the better.

rareddy:
Are you envisioning this separate from the view sql editor we have now? I feel that could be confusing to users two different ways to create view perhaps? I envisioned the SQL client being solely for querying the views. If not we need to see if we can bring them as one single offering

shawkins:
> Are you envisioning this separate from the view sql editor we have now?

Probably

> I feel that could be confusing to users two different ways to create view perhaps?

It's not confusing as that's the dominant working paradigm for anyone using a traditional sql client tool.

> I envisioned the SQL client being solely for querying the views. 

That's a different definition of sql client than what most expect.  I get that if we want to provide OData preview - then it needs to be at the view level and above (allow for expands etc.), but if you are going to offer a sql client it should provide the ability to perform any adhoc query.  If I'm creating a view and want to know what values are in some source table, do I have to go create a view, then preview, then go back to my current view editor panel?  Or should I be able to have a sql client where I can generally play around until I'm ready to create a view?  By all means what we have is straight-forward for the initial case of refining projection and filtering, but it makes anything more exploratory much harder than traditional tools.

rareddy:
> I get that if we want to provide OData preview

OData preview is not being considered AFAIK so far yet, we can consider it later I guess.

>Or should I be able to have a sql client where I can generally play around until I'm ready to create a view?

So the question is whether to gather the view name first then do ad-hoc query vs ad-hoc query then provide a name? We need to channel this into a single solution now, changing this later is only will create more confusion.

>By all means what we have is straight-forward for the initial case of refining projection and filtering, but it makes anything more exploratory much harder than traditional tools.

Is this towards the SQL Editor or the UI for view part you are referring to?

shawkins:
> OData preview is not being considered AFAIK so far yet, we can consider it later I guess.

I'm not proposing that.  I'm just saying that's the only rationale for a preview that restricts you to only the views that you've already created.  In our case since there is already a preview on the same page as the view editor, so the current sql client seems redundant.

> So the question is whether to gather the view name first then do ad-hoc query vs ad-hoc query then provide a name? 

No.  The dominant paradigm in most sql tools is just a sql client pane where you can execute whatever you want - whether you end up defining a view, procedure, etc. doesn't matter - you do / explore whatever you want.  If you have a concern about issuing updates, I'd get that - but you can always mark the sources as immutable.

> We need to channel this into a single solution now, changing this later is only will create more confusion.

My confusion comes from having a "sql client" that offer only limit/offset functionality (which we could have pagination automatically handled) over what's built-in to the editors.

So I'm saying ditch the current incarnation of the "sql client" and instead open it up to the user executing arbitrary sql against the dv specific preview vdb.  Whether we want to have that working state be persistent or not would be open to debate.  The more this could function like dbeaver, sqluirrel, eclipse data tools, etc. - the more time it will save someone in terms of development/exploration.  That looks like the ability to potentially multiple statement fragments and execute one at will.  Ideally from there if the user likes the sql there should be a short cut to create a view from that sql - but if that's a stretch then new view should support something like paste sql as an option rather than just being able to select a source table.

PR #217, opened 2019-09-03
TTOOLS-688 Finishes komodo changes for code complete

blafond:


shawkins:
LGTM, but there's a checkstyle failure:

src/main/java/org/komodo/rest/V1Constants.java:[279] (regexp) RegexpSingleline: Line has trailing spaces.

This is from the syndesis checkstyle.  They might have a way to automatically handle formatting issues.

blafond:
There was a Tab on that line which I removed... plus I added some  //$NON-NLS-1$ to some string constants for good measure.

PR #218, opened 2019-09-07
TEIIDTOOLS-760 TEIIDTOOLS-753

shawkins:
cleaning up embedded import handling,
async metadata loads,
adding an h2 source for testing
and changing the name of the main preview vdb


shawkins:
There are now four types of vdbs deployed to the embedded server:
1. Connection metadata loading vdbs, which have the suffix conn-load.  These are new and are for narrowing the window of time in which the preview vdb is invalid and/or missing source metadata.
2. Connection metadata vdbs, which have a conn suffix.  They are created only after the ddl has been captured by the load vdbs
3. The main preview vdb (just named Preview), which imports all of the connection metadata vdbs and is regenerated on every connection event
4. dv specific preview vdbs, which import the main preview vdb and get regenerated as view metadata changes or by any connection event

These changes are built up adding an async-load flag to teiid vdbs https://github.com/teiid/teiid/pull/1175 - so the Teiid snapshot needs updated first.  

That allows us to mark the loading vdbs as async such that the connection synchronizer thread won't be held up by importing metadata.  Rather than using polling, these changes add a VDBLifeCycleListener to react to those deployments, which can be to save the ddl, and then trigger the refresh of the downstream vdbs.

This addresses most of the potential concurrency issues, but does leave a narrow case - if during query execution the dv specific preview vdb is redeployed, the query can fail.  Guarding against that is rather expensive, so for now I'd just let the exception happen and the ui can resubmit.

rareddy:
what is the need for (3) vdb given (4)?

shawkins:
> what is the need for (3) vdb given (4)?

It primarily provides a single vdb to specify all the imports on.  Also if we assume that the master preview vdb is highly available, then it provides a way to not build the dv specific preview vdb as much - for views that don't reference other views, the dv specific preview vdb is not needed for validation nor execution (with some assumptions).  We can make an assessment of the best optimizations under https://issues.jboss.org/browse/TEIIDTOOLS-763

rareddy:
I am good for merging, looks like there a failing test

shawkins:
> I am good for merging, looks like there a failing test

It needs the updated snapshot.  I'll push that, and show this as clean.

shawkins:
I can't seem to make Travis happy.  I've locally pushed the Teiid snapshot several times.  I'll I'm getting is bad checksums in the travis build, even after clearing the cache.

shawkins:
I'm not sure what is going on.  Locally I can push a snapshot, but something's wrong with the checksums.  Updated Teiid to push the snapshot on every master commit, and it's still not working: https://travis-ci.org/teiid/teiid/jobs/582851311

shawkins:
Manually pushed common core again, now everything seems to work.  I'll back out the teiid master change if a snapshot on every commit not work.

PR #219, opened 2019-09-09
TEIIDTOOLS-764 adding validations to view listing and view definition

shawkins:
also correcting test package names


shawkins:
This has been updated to combine stash / validate and to return the full entity from stash.  As currently implemented it means that if the preview vdb is not available a service unavailable response will be returned rather than saving the entity.

rareddy:
LGTM

PR #221, opened 2019-09-10

shawkins:

### DIFF
--- a/server/komodo-core/src/main/java/org/komodo/WorkspaceManager.java
+++ b/server/komodo-core/src/main/java/org/komodo/WorkspaceManager.java
@@ -54,6 +54,8 @@ public interface WorkspaceManager {
 
     boolean deleteDataVirtualization(String virtualizationName);
 
+    List<String> findDataVirtualizationNames();
+
 
     ViewDefinition findViewDefinitionByNameIgnoreCase(String dvName, String viewName);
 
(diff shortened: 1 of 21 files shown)


shawkins:
This gets thing about halfway - it does not yet address making the dv name the view schema name, but it makes sure that the dv name is a valid open shift name and removes the pervasive references to vdb names. Looking at the syndesis generated image deployment config and associated, they keep a lot more information on theirs:

  template:
    metadata:
      annotations:
        openshift.io/deployment-config.latest-version: '1'
        openshift.io/deployment-config.name: i-my-name
        openshift.io/deployment.name: i-my-name-1
        prometheus.io/port: '9779'
        prometheus.io/scrape: 'true'
        syndesis.io/deployment-version: '1'
        syndesis.io/integration-id: i-LnxP7EQV6-Wp1Igyf4zz
        syndesis.io/integration-name: my name
      creationTimestamp: null
      labels:
        deployment: i-my-name-1
        deploymentconfig: i-my-name
        syndesis.io/app: syndesis
        syndesis.io/component: integration
        syndesis.io/deployment-version: '1'
        syndesis.io/integration: i-my-name
        syndesis.io/integration-id: i-LnxP7EQV6-Wp1Igyf4zz
        syndesis.io/type: integration
        syndesis.io/username: developer

For now rather than add more information, I just changed our side to go through the list of virturalizations when asking for all status so that we can associate the virtualization name.

shawkins:
> we been using vdb name and virtualization name interchangeably, need to check how this affects the service names etc.

Yes, this change makes the openshift name the base for all names in openshift.

PR #223, opened 2019-09-10

shawkins:


rareddy:
The one concern I have is user having to use "dv-foo" name in the JDBC/PG URL, when they just named as "foo"

PR #225, opened 2019-09-10

shawkins:


shawkins:
This affects the ui by removing the use of vdb names from service methods.

This makes the dv name the same as the published vdb name - the user will use the dv name in their connection strings.

It also ensures that the dv name is a valid openshift name, that is dv-lower case name can be used for an image, deployment name, etc.

This does not yet address making the dv name the view schema name, but I'll add that to this pr.

shawkins:
This is now up-to-date with master, and has TEIIDTOOLS-777 so that source stuff requires qualification and virtual stuff does not.  That affects the ui in two ways - the first is that preview queries against views should be unqualified as mentioned above.  The other is that auto complete or other references to source stuff does need to be qualified - which is a best practice regardless.  If that second change seems too strict, then we'd need a different approach to resolving, such as a schema search path - but that would need new logic on the Teiid side.

rareddy:
On the qualification side, I would error on what user would normally do, since this is a hand-coded editor, I imagine that would be unqualified?

shawkins:
> On the qualification side, I would error on what user would normally do, since this is a hand-coded editor, I imagine that would be unqualified?

Keep in mind you've always wanted preview queries of views to be unqualified...

There are issues with allowing unqualified source references - TEIIDTOOLS-754 when new sources become available, names can become ambiguous after the fact, names can conflict with view names, etc.  So this is really just enforcing a best practice that would make the sql more resilient without us having to do post processing to add qualification.  As long as auto-complete understands how to qualify things, then I'm fine with this requirement.

rareddy:
The auto-complete is doing qualified ones, but the initial generated query is not, maybe that is where change is needed. Also trying to think about how best to enforce this without becoming a hurdle in user activities. Like type in unqualified have a squiggly line or something. Auto-correction may end up wrong with multiple identical columns.

shawkins:
> The auto-complete is doing qualified ones, but the initial generated query is not, maybe that is where change is needed. 

There is probably a misunderstanding about what qualification I'm talking about here - this only affects table names.

rareddy:
Ok, I think those are already correct, and you are correct the auto-complete does not have the qualification with the source. That needs to be corrected irrespectively IMO. I will log a JIRA

rareddy:
logged https://issues.jboss.org/browse/ENTESB-11630

shawkins:
@mdrillin updated to correct the issue with the status.  

Eventually we'll likely need to pass or lookup the DataVirtualization object for the TeiidOpenShiftService as we'll likely want to associate more metadata/annotations (such as the uuid) and we may need to generate / save the openshift name on it.

rareddy:
Is this good for merging?

PR #226, opened 2019-09-10
ENTESB-10899

rareddy:


shawkins:
This looks good for the initial effort.  A couple of thoughts.  

What happens if the connection is in use when the delete is issued?  The syndesis ui won't let you remove an in use connection.  

It's also probably a good time to talk about security implications.  This is assuming a non-authenticated connection over unsecured jdbc.  When are we looking to hook our virtualizations up to keycloak?  Currently once a virtualization is published it's making everything exposed effectively public within the namespace.

Once backed by keycloak do you expect there to be a built-in service account for development/syndesis usage and/or would you want certificate based authentication over mms like was discussed for pg?

rareddy:
As per the delete is concerned, yes the UI will prohibit from deleting it if there is an Integration. But if the user is "unpublishing" the virtualization, the connection is already broken, thus I took the liberty to remove it. We need to addon checks like if the Virtualization is used in Integration to prohibit the user from "unpublish", then they will sync up.

Coming to security, I think Keycloak integration is at least quater or couple quarters away, as it will be available in RHMI but not in with Syndesis. Right now Integrations are also suffering this same issue. 
I would like to solve the certificate stuff ASAP(7.6), such that we have some default security starting with "DV on OpenShift", which will involve the Operator changes. In 7.6, I want to replace the "publish" mechanism with Operator in "teiid-syndesis", then that should back it here. Can you log an issue for certificate support in Operator? 

You can also add one for Keycloak, there we need to invent some kind of discovery mechanism by Operator. Last I heard the Operator for Keycoak was not fully ready. Until that time we have to relie on the manual configuration with the help of documentation.

shawkins:
> Can you log an issue for certificate support in Operator?

Logged https://issues.jboss.org/browse/ENTESB-11602

rareddy:
Good for merge then?

PR #229, opened 2019-09-17
ENTESB-10894, ENTESB-10896

rareddy:


rareddy:
BTW, this does not affect any of UI code.

server/komodo-server/src/main/java/org/komodo/openshift/TeiidOpenShiftClient.java
```diff
@@ -586,10 +581,8 @@ public void createDataSource(DefaultSyndesisDataSource scd) throws Exception {
         }
 
         //now that the komodoname is set, we can create the properties
-        Map<String, String> properties = scd.convertToDataSourceProperties();
-        properties.put(ID, scd.getId());
-
-        this.metadata.createDataSource(toUse, dsType, encryptionComponent.decrypt(properties));
+        TeiidDataSource teiidDS = scd.createDataSource(toUse);
```
shawkins: Maybe pass the DefaultSyndesisDataSource into a registerDataSource, then we can just get rid of the getDataSource check and handle everything in the DefaultMetadataInstance.

server/komodo-server/src/main/java/org/komodo/metadata/internal/DefaultMetadataInstance.java
```diff
@@ -353,14 +370,24 @@ public TeiidDataSourceImpl getDataSource(String name) throws KException {
     }
 
     @Override
-    public void deleteDataSource(String dsName) throws KException {
-        TeiidDataSourceImpl ds = this.datasources.remove(dsName);
-        if (ds != null) {
-            this.server.removeConnectionFactoryProvider(dsName);
+    public synchronized void deleteDataSource(String dsName) throws KException {
```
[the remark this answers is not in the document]
rareddy: corrected as suggested

server/komodo-server/src/main/java/org/komodo/openshift/TeiidOpenShiftClient.java
rareddy: corrected as suggested

rareddy:
@shawkins I have updated the code per your suggestions.

shawkins:
Merged this with additional changes to consolidate down to a single map of sources.  Also adding better handling of the openshift client in a follow-on commit to help with the integration test.

PR #230, opened 2019-09-19
ENTESB-10899: Adding JDBC connection in Syndesis connections when a v…

rareddy:
…irtualization is published. Also, making sure it is not unpublished when the connection to it is being used in a Integration.

There will be small follow on work on the UI to disable "unpublish" when the connection is used in an Integration.

### DIFF
--- a/server/komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionSynchronizer.java
+++ b/server/komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionSynchronizer.java
@@ -102,10 +102,12 @@ public class SyndesisConnectionSynchronizer {
 
     public void addConnection(DefaultSyndesisDataSource sds, boolean update) {
         try {
-            // this is avoid circular creation of the virtualization connection that is published through syndesis
-            DataVirtualization dv = kengine.getWorkspaceManager().findDataVirtualization(sds.getSyndesisName());
-            if (dv != null && dv.getId().equals(sds.getId())) {
-                return;
+            // this is avoid circular creation of the virtualization connection that is
+            // published through syndesis
+            for (DataVirtualization dv : kengine.getWorkspaceManager().findDataVirtualizations()) {
+                if (dv.getSourceId() != null && dv.getSourceId().equals(sds.getSyndesisConnectionId())) {
+                    return;
+                }
             }
         } catch(KException e) {
             LOGGER.warn("Error while adding a connection " + sds.getSyndesisName(), e);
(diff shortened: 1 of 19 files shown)


server/komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionSynchronizer.java
```diff
@@ -94,14 +94,23 @@ private void handleAddConnection(EventMsg event, boolean update) throws KExcepti
...
         if (sds != null) {
             deleteConnection(sds);
         }
     }
 
     public void addConnection(DefaultSyndesisDataSource sds, boolean update) {
+        try {
+            // this is avoid circular creation of the virtualization connection that is published through syndesis
+            DataVirtualization dv = kengine.getWorkspaceManager().findDataVirtualization(sds.getSyndesisName());
+            if (dv != null && dv.getId().equals(sds.getId())) {
```
shawkins: These are not the right checks.  You should not lookup the datavirtualization by the syndesisname because that may not be stable - it should be found by the source id matching the DefaultSyndesisDataSource id.

We should probably rename DefaultSyndesisDataSource.getId as getSyndesisId - as it is the syndesis assigned id.  

dv.getId on the other hand returns the teiid-komodo generated id, so an equality check won't work.

rareddy: yes, that was mistake, I started with corrected in other places, looks like I missed here.

rareddy:
@shawkins corrected and implemented your suggestions.

server/komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionSynchronizer.java
```diff
@@ -94,14 +94,25 @@ private void handleAddConnection(EventMsg event, boolean update) throws KExcepti
...
         if (sds != null) {
             deleteConnection(sds);
         }
     }
 
     public void addConnection(DefaultSyndesisDataSource sds, boolean update) {
+        try {
+            // this is avoid circular creation of the virtualization connection that is
+            // published through syndesis
+            for (DataVirtualization dv : kengine.getWorkspaceManager().findDataVirtualizations()) {
```
shawkins: This should be replaced by a new repository method - findDataVirtualizationBySourceId, or a boolean function.

rareddy:
continued in pull https://github.com/teiid/teiid-syndesis/pull/232

PR #231, opened 2019-09-19
TEIIDTOOLS-743 flattening the project structure

shawkins:


rareddy:
LGTM. Removing the "server" and making "core" and "server" at the root level.

PR #242, opened 2019-09-27
TEIIDTOOLS-728 initial single virtualization export

shawkins:
Here is the single data virtualization import/export.  It uses:

get .../dataservices/name/export
post .../dataservices/import?virtualization=name

The import/export format is a zip file containing a dv.json and a dv-info.json file - which is basically the same approach as syndesis.  We don't yet need any handling of versioning beyond the usage of the adapter classes which means that we can change the other service methods and/or the database and not disturb the import/export format.

The get method does seem problematic - as described in another issue - it seems like gets under an resource path should be for that object or child information.  So it could just as easily be a post or moved to it's own service (I recall that komodo had an import/export service).

This does not attempt to resolve any differences in the declared sources and what exists - it will just note the differences in the status returned by import.  At worst the imported virtualization will just have a bunch of errors.  We may eventually need to update the sql if the name of a given connection has changed between environments.  Another simplifying assumption is that everything gets a new uuid, rather than only creating on conflict.

Before we implement the bulk import/export, we need to add the labeling functionality.

This also contains a small fix of the 404 for getDataservice.

### DIFF
--- a/komodo-server/src/main/java/org/komodo/metadata/TeiidDataSource.java
+++ b/komodo-server/src/main/java/org/komodo/metadata/TeiidDataSource.java
@@ -23,7 +23,7 @@ public interface TeiidDataSource {
 
     String getTranslatorName();
 
-    String getId();
+    String getSyndesisId();
 
     Object getConnectionFactory();
 
--- a/komodo-server/src/main/java/org/komodo/metadata/internal/TeiidDataSourceImpl.java
+++ b/komodo-server/src/main/java/org/komodo/metadata/internal/TeiidDataSourceImpl.java
@@ -90,7 +90,7 @@ public class TeiidDataSourceImpl implements Comparable<TeiidDataSourceImpl>, Tei
     }
 
     @Override
-    public String getId() {
+    public String getSyndesisId() {
         return this.id;
     }
 
--- a/komodo-server/src/main/java/org/komodo/openshift/TeiidOpenShiftClient.java
+++ b/komodo-server/src/main/java/org/komodo/openshift/TeiidOpenShiftClient.java
@@ -17,33 +17,11 @@
  */
 package org.komodo.openshift;
 
-import java.io.BufferedReader;
-import java.io.BufferedWriter;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.PrintWriter;
-import java.io.Reader;
+import java.io.*;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.Calendar;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
-import java.util.WeakHashMap;
+import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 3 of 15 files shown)


rareddy:
Looks good overall. Simple.

I am kind of surprised at first glance at the approach. As I would first thing I would think about is exporting the VDB in DDL form. But the intention behind this import/export is purely for migration between environments, then it makes sense. I sure hope no one puts this in GIT and ask us to support this the VDB format, that would be really BAD. I would probably capture Syndesis version and the DB schema version to analyze any version differences if we get into a backward compatible situation.

shawkins:
> I sure hope no one puts this in GIT and ask us to support this the VDB format, that would be really BAD.

We don't really care what they do with the export - but no this isn't a VDB - it's the state of the teiid-syndesis virtualization.

>  I would probably capture Syndesis version and the DB schema version to analyze any version differences if we get into a backward compatible situation.

Neither are needed as we are including our own version file in the zip - see dv-info.json

> BTW, I did not understand the requirement Adapters, I would have thought regular entity with json annotations would have done the job?

That would effectively be the same thing.  The adapter sits on top of the actual entity and gives us a consistent import/export format so that and db concern or client/server concern won't affect this.

rareddy:
>Neither are needed as we are including our own version file in the zip - see dv-info.json
I see that IMO that is more burden on us to keep this version along with any changes to db schema version and their compatibility + if there are any additionals like souces or extension jars etc. it is more straight forward to keep schema versions then we can just say something backward compatible or not. Eventually, we will have to write migration between versions, unfortunately :( At that time we probably we would feel DDL would have been better :)

>That would effectively be the same thing.
Here we are adding a new wrapper class for every entity, and they change when entity changes right? Are these annotations somehow interfering with the entity persistence into db? For me, it appeared as verbose, but since we do not have many entities right now it is manageable.

shawkins:
> I see that IMO that is more burden on us to keep this version along with any changes to db schema version and their compatibility

You seem to be conflating several concerns.  The db has it's own needs and even some usage of json serialization - some of that is derived state that we don't need in the export format.  Also any change that simply adds or removes a field won't necessitate a change in the import/export version.

> Eventually, we will have to write migration between versions, unfortunately :( At that time we probably we would feel DDL would have been better :)

I don't follow you.  Are you talking about teiid-syndesis version migration where the migration is applied via flyway and related logic, or are you talking about version changes between older exports and newer teiid-syndesis versions?

> Here we are adding a new wrapper class for every entity, and they change when entity changes right?

Define change.  If the entity changes the name of a field, yes the adapter has a different getter and setter to call, but it's import/export contract stays the same.

> Are these annotations somehow interfering with the entity persistence into db? For me, it appeared as verbose, but since we do not have many entities right now it is manageable.

It's best to formalize the concerns.  There is a core entity that is persistent - which may or may not utilize json serialization on it's own (currently only view definition does for the extended state so that we don't have to have a fixed schema).  You have presentational views of the core entities, which sometimes mix in information - note that jackson does have the concept of json views that we aren't currently utilizing which can replace some of the RestXXX classes if you want.  Finally the import/export format is all about stability - once we bless version it should be supported for quite some time.  I don't want to have to update this support for every presentational or storage level change.

rareddy:
>Also any change that simply adds or removes a field won't necessitate a change in the import/export version.

I thought this would be a change to import/export, actually, any schema change that is not compatible with previous version will change the import.

> are you talking about version changes between older exports and newer teiid-syndesis versions?

yes

> Define change.

Like adding a column, or a new entity. Then adapter classes need to change.

I do not think I follow you RestXXXX comment, as I do not know jackson json views.

> Finally the import/export format is all about stability - once we bless version it should be supported for quite some time.

exactly. Atleast my assumption is the export format is only compatible when the schema between versions of the environments are compatible, otherwise not. But you are trying to tell otherwise, that is where I am confused :(

shawkins:
> I thought this would be a change to import/export, actually, any schema change that is not compatible with previous version will change the import.

Which schema are you referring to?  The adapters provide a consistent import/export format such that you isolate changes as much as possible.  You only introduce a new adapter on a structural change or property renaming.

> Like adding a column, or a new entity. Then adapter classes need to change.

Again define change.  If you add a new column that is exposed to export that won't matter to the existing versions - note that the ObjectMapper used for import has ignores missing properties, so you can have all the extra stuff in there that you want and it won't matter.

The same reasoning with a new entity - as long as it doesn't change the structure of existing entities you can add that and it won't matter to the prior versions.

> exactly. Atleast my assumption is the export format is only compatible when the schema between versions of the environments are compatible, otherwise not. But you are trying to tell otherwise, that is where I am confused :(

Again which schema are you referring to?  The adapters define a consistent json import/export schema that allows for new/missing stuff - only when you rename or change the entity structure do you need a new adapter.

rareddy:
> Again which schema are you referring to? 
The scheme that is defined through Flyway to PostgresDB to store the details of virtualizations. I do not know of any other schemas that we support.

> only when you rename or change the entity structure do you need a new adapter.

All I am asking if there is change like this between two different syndesis versions of the export, are you saying it will work without any issues because of adapters or somehow where version number semantics is going to be used either reject the import based on some check? 

At the end of the day, the way I see this as export of database table(s) data for respective virtualization, then import on another end. The DB schema for between these are not compatible then we end up in error during the import. I am not asking we should support that, but if/how we need to detect that and prevent or reject it.

shawkins:
> All I am asking if there is change like this between two different syndesis versions of the export, are you saying it will work without any issues because of adapters or somehow where version number semantics is going to be used either reject the import based on some check?

Short answer is yes, this approach will work.  Longer answer:  It matters what the change is as to whether you need to do anything with the adapters.  If you make a database level change to the derived state (tracking more information about columns dependencies for example), then no change is needed at all to the adapters.  If you rename a database level field, you just change the getter/setter the adapter is referencing - there is no change in the import/export version.  So on and so forth.  

When there is change to what is exported:  if it is just a new field, then you still don't have to do anything most of the time.  Exporting from a newer version that includes the field and importing to an older instance that does not will not fail as the objectmapper will ignore missing fields.  So as long as you can handle the null case in the other direction you don't need any special version related logic.  Only if you make some sort of rename or structural change in what is actually exported do you need to consider a new adapter.  You still leave the old adapter in place to handle imports of older compatible versions.  

Keep in mind that there is no built-in support for json schema versioning in jackson.  It's up to the user to decide upon a strategy for schema migration.  You can wire in lower level jackson resolvers, potentially use views, custom extensions, or adapters.

> At the end of the day, the way I see this as export of database table(s) data for respective virtualization, then import on another end. The DB schema for between these are not compatible then we end up in error during the import. I am not asking we should support that, but if/how we need to detect that and prevent or reject it.

An adapter provides stability between your database schema and your json import/export schema.  As long as the import contains sufficient information (meaning that a lot of derived / optional state can be missing), then the import can proceed.

PR #243, opened 2019-09-30
TEIIDTOOLS-789 TEIIDTOOLS-729

shawkins:
adding a timer job for refresh and cleaning up removed

### DIFF
--- a/komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionMonitor.java
+++ b/komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionMonitor.java
@@ -25,6 +25,7 @@ import javax.annotation.PostConstruct;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.komodo.KException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
@@ -189,7 +190,7 @@ public class SyndesisConnectionMonitor {
                 Message msg = mapper.readValue(text.getBytes(), Message.class);
                 if (msg.getEvent().contentEquals("message") && msg.getData().contentEquals("connected")) {
                     connected = true;
-                    connectionSynchronizer.synchronizeConnections();
+                    connectionSynchronizer.synchronizeConnections(false);
                 } else if (msg.getEvent().contentEquals("change-event")) {
                     EventMsg event = mapper.readValue(msg.getData().getBytes(), EventMsg.class);
                     if (event.getKind().contentEquals("connection")) {
@@ -219,5 +220,14 @@ public class SyndesisConnectionMonitor {
     @PostConstruct
     public void init() {
         this.executor.scheduleAtFixedRate(()->this.connect(), 5, 15, TimeUnit.SECONDS);
+        this.executor.scheduleAtFixedRate(()->{
+            try {
+                if (connected) {
+                    connectionSynchronizer.synchronizeConnections(true);
+                }
+            } catch (KException e) {
+                LOGGER.error("failed to synchronize", e);
+            }
+        }, 5, 5, TimeUnit.MINUTES);
     }
 }
--- a/komodo-server/src/main/java/org/komodo/openshift/TeiidOpenShiftClient.java
+++ b/komodo-server/src/main/java/org/komodo/openshift/TeiidOpenShiftClient.java
@@ -17,16 +17,7 @@
  */
 package org.komodo.openshift;
 
-import java.io.BufferedReader;
-import java.io.BufferedWriter;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileWriter;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.PrintWriter;
-import java.io.Reader;
+import java.io.*;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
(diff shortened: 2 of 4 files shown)


shawkins:
This adds a 5 minute refresh job.  This is not ideal for a multi-pod environment - there there is already an existing issue to address event distribution (ideally we'd only perform the metadata load from a single member as well).

komodo-server/src/main/java/org/komodo/rest/connections/SyndesisConnectionMonitor.java
```diff
@@ -219,5 +220,14 @@ public void close() {
     @PostConstruct
     public void init() {
         this.executor.scheduleAtFixedRate(()->this.connect(), 5, 15, TimeUnit.SECONDS);
+        this.executor.scheduleAtFixedRate(()->{
```
rareddy: Can we make this configurable?, 5 mins seems too short for me. Also, we need a way to turn it off.

shawkins: If you are waiting for a table to show up, 5 minutes is almost certainly too long :)

Not that I'm opposed to adding a point of configuration, but:

Who would configure it?

And under what circumstances would you turn it off?

rareddy: we can configure or shutoff using the ENV variable. I still do not understand why we can not be explicit from UI, although it is under a given VDB, what will be the issues with invalidating views from multiple VDBs if needed? in production system loading the metadata every 5 mins seems overkill, this is more needed in development time IMO, we need a way to control this behavior or do it as explicit action.

shawkins: > I still do not understand why we can not be explicit from UI, although it is under a given VDB, what will be the issues with invalidating views from multiple VDBs if needed?

I don't quite follow you here.  It's not under a specific vdb - this refreshes the metadata for a connection which is effectively global to the teiid-syndesis instance. If the source metadata changes, then yes some of your views can be invalidated - but that is expected/desirable.

> in production system loading the metadata every 5 mins seems overkill, this is more needed in development time IMO, we need a way to control this behavior or do it as explicit action.

This is approximately the behavior that data connect exhibited - no user action for refresh, but periodic.  If you want to suggest a different default that's fine, but making it too long is effectively useless from a usability perspective.  As far as the type of environment that teiid-syndesis is deployed in that would be something new to capture, but realistically what you'd rather have here is something driven by cdc.

You could also make a case for this to be part of some kind of validate/refresh action on a given virtualization if you really hate the background task.

rareddy: CDC would be optimal but given vast amount of sources I do not think we will ever get there any time soon.

>You could also make a case for this to be part of some kind of validate/refresh action on a given virtualization if you really hate the background task.

IMO, whether it is 5 min or 50 minutes if user is not known about it could be still an issue, with an explicit refresh, we can handle user expectation as well as unnecessary metadata loads. I definitely would prefer this model. If we need to do this on connection panel with the update there we can negotiate that with Syndesis UI, otherwise, we can consider on table selection window.

shawkins: > IMO, whether it is 5 min or 50 minutes if user is not known about it could be still an issue, with an explicit refresh, we can handle user expectation as well as unnecessary metadata loads. 

A couple of things.

1. Doing the background approach is a workaround for not having the explicit action.  Some of the confusion is probably because you weren't on the morning standup call. The https://issues.jboss.org/browse/ENTESB-11137 issue still apparently has an expectation of being addressed in the short term.  Since there wasn't a clear path forward to do that with the ui, the decision was to do it with the periodic refresh.
2. Unnecessary metadata loads seems like a minimal issue.

rareddy: I think we are going in circles here :) I still do not a good answer why we think there is no clear path from UI? If needs to be we can push this to 7.6, for resolution. But, I disagree that full metadata load every 5 mins is a minimal issue, especially if we are talking to large databases, salesforce etc.

shawkins: > If needs to be we can push this to 7.6, for resolution.

That is entirely the point - it cannot simply be pushed to 7.6.  This is considered a very high priority usability issue for 7.5.

> I still do not a good answer why we think there is no clear path from UI?

This has been discussed several times.  The determination is that any contribution to the syndesis side won't land in 7.5.  There is also not a great place for this under our virtualization centric UI.  The closest would be something like a kebob item on a virtualization card that to refresh all source metadata - with appropriate language that the refresh will affect all virtualizations.

> But, I disagree that full metadata load every 5 mins is a minimal issue, especially if we are talking to large databases, salesforce etc.

I don't quite follow what the concern is there - especially from a user facing perspective which is the primary concern of a tech preview.  There is plenty about our images (cpu/memory) and our teiid-syndesis server whole approach to metadata that will have scalability and other other concerns - it seems odd to be so focused on this.

> I think we are going in circles here :)

Again, if you want to suggest a different default or push for a UI contribution location that we control that's great.  If you just say I don't like it, then yes we are stuck.

rareddy: > here is plenty about our images (cpu/memory) and our teiid-syndesis server whole approach to metadata that will have scalability and other other concerns
Here the concerns are around Syndesis, not on the external systems. With the change, it will be a burden on external systems, and we know 99.9% of times there is no updates to the metadata.

> The determination is that any contribution to the syndesis side won't land in 7.5.
Are you suggesting this is only for 7.5, then I am fine with it.

> if you want to suggest a different default or push for a UI contribution location that we control that's great
This can be done in the with refresh button under "create view"/"import data source" IMO. That seems to be the logical place. We can run it by the UX to get buy-in. As I mentioned above we can move in this direction in 7.6

rareddy: I agree with both places in UI make sense.

Ok, let's resolve this as is here, then keep the ENTESB-11137 open with partial completion comment and may be with comments on UI updates for 7.6.

rareddy:
Steve, can you also submit a PR for product branch pls.

PR #256, opened 2019-10-07
TEIIDTOOLS-737 deprecating the validation methods

shawkins:
This makes the validation change mentioned on TEIIDTOOLS-737.  Once the ui is changed, we can then remove the KomodoVdbService.

### DIFF
--- a/komodo-core/src/main/java/org/komodo/WorkspaceManager.java
+++ b/komodo-core/src/main/java/org/komodo/WorkspaceManager.java
@@ -48,6 +48,8 @@ public interface WorkspaceManager {
 
     DataVirtualization findDataVirtualization(String virtualizationName);
 
+    DataVirtualization findDataVirtualizationByNameIgnoreCase(String virtualizationName);
+
     DataVirtualization findDataVirtualizationBySourceId(String sourceId);
 
     public Iterable<? extends DataVirtualization> findDataVirtualizations();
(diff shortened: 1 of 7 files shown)


rareddy:
LGTM - Should we log another issue to remove the deprecation and fix the UI side?

mdrillin:
> LGTM - Should we log another issue to remove the deprecation and fix the UI side?

No need to log another issue for UI work - I can work the UI side under TTOOLS-737.

rareddy:
:+1: Please also remove deprecated calls in that commit.

shawkins:
I should add that there is one small difference in that this does not validate virtualization names against connection names as the validate method did.  Would you be ok with it returning a 303 "see other" code if the name matches a source name?

mdrillin:
I'm ok with the 303 code.  I'm not clear atm how to validate view names.  What is service call for that?

shawkins:
> I'm ok with the 303 code. I'm not clear atm how to validate view names. What is service call for that?

This adds GET dataservices/**name**/views/**name** - if you call that and get anything other than a 404, you'll know that the name is invalid (forbidden), an existing data virtualzation, or a connenction name (see other).

mdrillin:
I finished the UI side changes to remove the nameValidation service calls and adapted to the GET service calls.  https://github.com/syndesisio/syndesis/pull/6875
Will merge once the UI PR is approved and ready to merge.

PR #257, opened 2019-10-09
TEIIDTOOLS-737 removing deprecated methods

shawkins:
now the that the ui has been updated, removing the old methods


shawkins:
I'm not sure if we want to do the deprecation then removal as a matter of practice for minor master changes.  It seems like that may allow for the commits to go in sooner, but if we're able to work things in tandem then it's unnecessary.  What do you guys think?  And are we ready for the larger change of cleaning up the endpoints and field names?

mdrillin:
Yeah, this is probably a good time to make the endpoint and naming changes.  I think it's ok to work simultaneously and just remove instead of deprecate, at least for now.

rareddy:
my feeling on deprecations is if it is a public api that used by more than just us, then deprecation will work well, but if it is just us then clear it away. I know this is not what might not want do, that my pragmatic view :)

PR #258, opened 2019-10-10
TEIIDTOOLS-737 removing the unused publishing methods

shawkins:


shawkins:
This removes the unused methods identified by Ramesh.  The next step would be to combine the publishing service and the data service service - and to update the end points in the process.  Does that sound good?

shawkins:
Also the main publish method may need some tweaking.  There's a lot going on there under a single transaction that will be holding a connection.

mdrillin:
> This removes the unused methods identified by Ramesh. The next step would be to combine the publishing service and the data service service - and to update the end points in the process. Does that sound good?

Yeah, that's fine with me.

PR #259, opened 2019-10-11
TEIIDTOOLS-737 simplifying and refactoring endpoints

shawkins:


shawkins:
This focuses only on endpoints - there are no changes to rest entity field names.

There are now three top level end points: virtualizations (was workspace/dataservices), metadata, and editors (was service/userProfile/viewEditorState)

Changes:

* Combined the publishing service into the data virtualization service, so instead of being under the metadata end point they are now under virtualizations
* Removed the name from the path of create data virtualization - instead of POST virtualizations/name, use just POST virtualizations
* Removed the view validation end point as it's been combined with save
* view listings are now at GET virtualizations/{name}/views - and is consistent with the single view / validation GET virtualizations/{name}/views/{name} in that they both now return a ViewListing.
* Removed service/about
* komodoSourceName is now teiidSourceName

Also do we want to leave the app version at v1?

mdrillin:
> Also do we want to leave the app version at v1?

I don't have a strong preference, we could probably leave it at v1 until we are production.

Minor issue, I noticed that {{teiidSourceName}} have double braces on the swagger page.  I should have the UI side changes ready to PR Monday...

shawkins:
Corrected the placeholder.

PR #263, opened 2019-10-22
Teiidtools 743

shawkins:


shawkins:
This is the preliminary refactoring prior to the repackaging.  This eliminates more unused stuff, and consolidates things to a single module.

shawkins:
I have the changes for the repackaging ready.  However to fully remove our usage of komodo it also:

changes the coordinate group to io.syndesis.dv and changes the database properties to dv. rather than komodo.connectionUrl.  Both of these require changes on the syndesis side.  Let me know if you want this to go in first before including those.

rareddy:
@shawkins did you mean on the UI side? may be @elvisisking can help us out in making those?

shawkins:
@rareddy on the syndesis side.  The coordinate change is just a permission issue with pushing snapshots/releases until we're merged, but for the properties there is:

install/operator/pkg/generator/assets/addons/komodo/addon-komodo-server.yml.tmpl:
`value: "-Djava.net.preferIPv4Stack=true -Duser.home=/tmp -Djava.net.preferIPv4Addresses=true -Dlog4j.logger.org.apache.http=DEBUG -Dkomodo.user={{.Syndesis.Spec.Components.Db.User}} -Dkomodo.password={{.Env.POSTGRESQL_PASSWORD}} -Dkomodo.connectionUrl=jdbc:{{.Env.POSTGRES_URL}} -Dkomodo.binaryStoreUrl=jdbc:{{.Env.POSTGRES_URL}} -Dkomodo.connectionDriver=org.postgresql.Driver"`

We can change our side to directly reference the url and password from the env, but it looks like we still need to pass in the db user.

There are also another 60+ references to the term komodo in syndesis source.

And are we keeping the vdb-builder term?  Seems like we can replace that with dv or syndesis-dv as well.

rareddy:
@shawkins ok, I believe they are also on ENV properties so if we adopt to those we can eliminate the db properties? The rest either I or yourself can submit the changes IMO. Let me know.

shawkins:
> @shawkins ok, I believe they are also on ENV properties so if we adopt to those we can eliminate the db properties?

That's what I'm saying "We can change our side to directly reference the url and password from the env, but it looks like we still need to pass in the db user." - the user can come from .Syndesis.Spec.Components.Db.User so we still need at least that property.

shawkins:
This is now up-to-date with all proposed changes.  The komodo term has been fully removed (there's still KException and KLog - which people will assume refers to https://archer.fandom.com/wiki/Kenny_Loggins) 

We're consistently using io.syndesis.dv and "dv" as the application name.  The full term DataVirtualization refers to the actual object.  Some of the packages have been combined and where easy we're using the syndesis term model, server, endpoint.  But not all class names have been updated.

shawkins:
On the syndesis side we have vdb-builder/komodo references in app/ui-react/syndesis/tapes/data-virtualization.spec.js/vdb-builder/v1/... - however these are still reference pre-refactoring endpoints, so I'm not sure how valid they are.

Here's what that pr will basically look like: https://github.com/shawkins/syndesis/commit/7e70983b46395610ee0d14280ce5f3c408062994 - this does not include the regeneration of the asset stuff.

rareddy:
I am guessing "app/ui-react/syndesis/tapes/data-virtualization.spec.js/vdb-builder/v1/... " are part of testing framework. You can also add in the change to couple config files looks like. Also the assets file they now have a post-commit hook to generate and commit. Otherwise looks good to me.

shawkins:
> I am guessing "app/ui-react/syndesis/tapes/data-virtualization.spec.js/vdb-builder/v1/... " are part of testing framework.

Yes, I'm saying it looks like they need regenerated regardless, so I didn't modify them.

> You can also add in the change to couple config files looks like.

What are you referring to?

rareddy:
config.xxx.json files at below location, 
https://github.com/syndesisio/syndesis/tree/master/app/ui-react/syndesis

also in file WithConfig.tsx file

shawkins:
Yes, there were dangling vdb-builder references in app/ui-react/syndesis and install/operator/pkg/generator/assets - I had replaced those in an earlier change, but didn't carry them forward for some reason.  Should be good now.

rareddy:
oh, ok. I am good with submit the commit.

shawkins:
> BTW, with this change, we will not able to cut release anymore ourselves, as we can only push "org.teiid" to maven central

Yes, I mentioned that up above.  Thinking about it more - it's probably best to back just that change out let @mdrillin and others approve this and do the merge on our side first.  Then we can tackle the next steps.

PR #266, opened 2019-11-15
Teiidtools 818

shawkins:


shawkins:
This is an initial implementing of publish versioning built on top of our existing logic for publish and export.  The biggest issue is that starting an old virtualization requires a full rebuild.  Another is that upon revert the uuids of restored views are changed - although that is something could be changed if needed.

PR #268, opened 2019-11-18
TEIIDTOOLS-840 using the working state metadata as the design time

shawkins:


shawkins:
Thinking about this more, that approach alone won't work as we'll have the same issue at runtime.  The idea is that we should keep the virtualization name out of the sql as much as possible as it will simplify any renaming, so it's best to allow the view name reference to be unqualified.  I see if there is something that makes sense to add at a lower level.

PR #271, opened 2019-11-19
[TEIIDTOOLS-842] - added javax.serlet dependency

vhalbert:


shawkins:
This is coming from tomcat-embedded-core

vhalbert:
Does that mean you are removing tomcat dependency or that an excludes needs to be added?

shawkins:
> Does that mean you are removing tomcat dependency or that an excludes needs to be added?

Neither.  What you are seeing is that the current pom has the basepom dependency checks disabled, and a local change has them enabled.  With them enabled they will warn/error on things like using a transitive dependency without listing explicitly as a dependency in your pom.  If you see where javax.servlet is currently coming from transitively it's tomcat-embedded-core via the spring boot web starter.  So to satisfy basepom we need to make that an explicit dependency - however since the version is already defined by the spring boot bom, we shouldn't need to specify the version.

vhalbert:
All the issues you see are from running the dv build withing syndesis, referencing the syndesis parent pom.

vhalbert:
I'm just backporting and building again in teiid so that we get these done before the porting of the code.

PR #273, opened 2019-11-19
[TEIIDTOOLS-845] Adding apache commons-logging dependency

vhalbert:


shawkins:
On this and all of the other used, but undeclared dependencies, check to see what is in the dependency tree.  In this case it's actually spring-jcl.  Also an explicit version won't be needed - that should be coming from another pom.

PR #274, opened 2019-11-19
[TEIIDTOOLS-842]  Updating classes based on the change on using syndesis parent pom

vhalbert:
Working thru the changes due to changing to use the syndesis as the parent pom.   This is not the last of the changes, only the next set that's been resolved.


shawkins:
Van, I don't see your pom change in any of the prs.  Are you simply removing the checkstyle and some of the basepom property overrides when switching to their parent pom?

vhalbert:
There's a combination of changes, as I move thru the next issue that pop's up.   The dependencies changes,  I think were fixed in the initial PR.  The subsequent changes appear to be class changes based on the checkstyle issue.   But until I get to the end, I'm not sure what issue I still may hit.

PR #275, opened 2019-11-20
[TEIIDTOOLS-842] Changes to support using syndesis parent pom

vhalbert:


shawkins:
merged into a later pr

